PR #2, opened 2019-09-20
Fix CI and cmake

vvolkl:
@graeme-a-stewart  it seems I had a typo in the previous PR, it has to be `.travis.yml` not `.travis.yaml`, now let's see if it runs!


init.sh
```diff
@@ -3,7 +3,7 @@ TOOLSPATH=/cvmfs/fcc.cern.ch/sw/0.8.3/tools/
 OS=`python $TOOLSPATH/hsf_get_platform.py --get os`
 
 # source FCC externals
-source /cvmfs/fcc.cern.ch/sw/views/releases/externals/96b.0.0/x86_64-$OS-gcc8-opt/setup.sh
+source /cvmfs/fcc.cern.ch/sw/views/releases/externals/94.2.0/x86_64-$OS-gcc62-opt/setup.sh
```
graeme-a-stewart: I guess we can bear this for the moment, but we should be more flexible about dependencies and not hard code paths

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - docker
+
+language: cpp
+
+env:
+  matrix:
+    - COMPILER=gcc
+
+before_install:
+  - wget https://gitlab.cern.ch/lhcb-core/LbDocker/raw/master/scripts/lb-docker-run
```
graeme-a-stewart: Seems very overengineered just to get a few options passed to the `docker run` command...

PR #3, opened 2019-09-25
First Test Converting From HepMC to Preliminary MCParticle

vvolkl:
Adds a quick test how to write the content of a HepMC::GenEvent to a EDM4HEP ROOT file.

Still to do: parent/sibling links

Some issues I noticed while writing the example:
* It would be nice if `setMomentum` could be called with three floats directly, instead of a FloatThree
https://github.com/vvolkl/EDM4HEP/blob/feature-hepmc/test/hepmc/edm4hep_testhepmc.cc#L118
* The output in the root file doesn't look at all like the one from the FCC PodioEventSvc implementation, which explicitly lists the members of the particle as branches. I'll have to check if I made a mistake with the podio writer.


test/hepmc/edm4hep_testhepmc.cc
```diff
@@ -0,0 +1,130 @@
...
+      std::cout << "Converting particle with Pdg_ID " <<  (*particle_i)->pdg_id() << " ..." << std::endl;
+      auto part = edm4hep::MCParticle();
+      part.setPDG((*particle_i)->pdg_id());
+      part.setGeneratorStatus((*particle_i)->status());
+      // look up charge from pdg_id
+      HepPDT::ParticleID particleID((*particle_i)->pdg_id());
+      part.setCharge(particleID.charge());
+      //  convert momentum
+      auto tmp = (*particle_i)->momentum();
+      part.setMomentum(edm4hep::FloatThree(tmp.px(), tmp.py(), tmp.pz()));
```
gaede: ... you could also write: `part.setMomentum( { tmp.px(), tmp.py(), tmp.pz() } );` which is close to a c'tor that takes three floats ...

vvolkl: Thanks Frank, that's definitely better! I ran into some issues (#8), but will update as soon as they are fixed

test/hepmc/edm4hep_testhepmc.cc
```diff
@@ -0,0 +1,130 @@
...
+      // contained in those vertices
+
+
+
+  //Part2: Convert the particles and write to file
+    auto store = podio::EventStore();
+    auto writer = podio::ROOTWriter("edm4hep_testhepmc.root", &store);
+    auto& particles       = store.create<edm4hep::MCParticleCollection>("testparticles");
+    writer.registerForWrite("testparticles");
+    for (auto particle_i = evt->particles_begin(); particle_i != evt->particles_end(); ++particle_i) {
```
andresailer: is 
```suggestion
    for (auto& particle: particles) {
```
Also possible?

vvolkl: Hi Andre, I was looking for that as well, but unfortunately HepMC doesn't use a standard container but some kind of self-made iterator, as you mention: http://lcgapp.cern.ch/project/simu/HepMC/205/html/classHepMC_1_1GenEvent.html#37646bf8f35efdcad9e9cb99d4afcf52
If you find  a nicer way let me know!

vvolkl:
I will still have to rebase on Frank's changes, and actually add some test if the parent/daughter links are working correctly

test/hepmc/edm4hep_testhepmc.cc
gaede: We have the same issue in PODIO: our iterators for relations and vector members are also not standard c++,e.g.:
```
  void addParent(edm4hep::ConstMCParticle);
  unsigned int parents_size() const;
  edm4hep::ConstMCParticle getParents(unsigned int) const;
  std::vector<edm4hep::ConstMCParticle>::const_iterator parents_begin() const;
  std::vector<edm4hep::ConstMCParticle>::const_iterator parents_end() const;
```
In plcio  have introduced a small helper vector_view that can be used in range based for loops: https://stash.desy.de/projects/IL/repos/plcio/browse/source/include/vector_view.h
with this and 
```
Extracode:
vector_view<ConstMCParticle> getParents() const { return vector_view<ConstMCParticle>( parents_begin(), parents_end()) ; } \n
```
one can write range based for loops:
```
for(auto d : mcp.getParents() ){ ...} 
```

If you think that would be useful, we could add it to PODIO as a feature...

andresailer: I think that would be very nice to have

PR #6, opened 2019-09-25
Add SimTrackerHit

vvolkl:
First Iteration: Taken from PLCIO, but with one long long int member for cellID instead of two ints, and without Extracode.

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -112,3 +112,18 @@ datatypes :
       void setOverlay(bool bitval) { setSimulatorStatus( set_bit( getSimulatorStatus() , BITOverlay , bitval ) ) ;   }         \n 
       "
       implementation : ""
+
+
+  edm4hep::SimTrackerHit:
+    Description: "Simulated tracker hit"
+    Author : "F.Gaede, DESY"
+    Members:
+      - long long int cellID          //ID of the sensor that created this hit
+      - float EDep           //energy deposited in the hit [GeV].
+      - float time           //proper time of the hit in the lab frame in [ns].
+      - float pathLength     //path length of the particle in the sensitive material that resulted in this hit.
+      - int   quality        //quality bit flag. 
+      - edm4hep::Vector3f position //the hit position in [mm].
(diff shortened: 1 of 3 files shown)


vvolkl:
One question: For FCC this class was one of the bottlenecks and took a lot of diskspace. The MCParticle linked to the Hit also has infomation on "momentum", but not at the point where the Hit was created. Furthermore some Digitization algorithms depend on the angle of the particle path crossing the sensor, which this also provides, so I think this should be kept as is, no?

gaede:
@vvolkl I had tried to make your PR compatible for rebasing but seem to have made a mistake.
I will simply create a new PR with Sim hits that we can include and then iterate on.

edm4hep.yaml
```diff
@@ -5,61 +5,125 @@ options:
...
+  edm4hep::SimTrackerHit:
+    Description: "Simulated tracker hit"
+    Author : "F.Gaede, DESY"
+    Members:
+      - long long int cellID          //ID of the sensor that created this hit
+      - float EDep           //energy deposited in the hit [GeV].
+      - float time           //proper time of the hit in the lab frame in [ns].
+      - float pathLength     //path length of the particle in the sensitive material that resulted in this hit.
+      - int   quality        //quality bit flag.
+      - plcio::DoubleThree position //the hit position in [mm].
```
andresailer:
```suggestion
      - edm4hep::Vector3d position //the hit position in [mm].
```

edm4hep.yaml
```diff
@@ -5,61 +5,125 @@ options:
...
+      - plcio::FloatThree momentum  //the 3-momentum of the particle at the hits position in [GeV]
```
andresailer:
```suggestion
      - edm4hep::Vector3f momentum  //the 3-momentum of the particle at the hits position in [GeV]
```

vvolkl:
Thanks for the comments, should be ok now except that it needs some changes in podio (https://github.com/AIDASoft/podio/pull/60) but of course you can also create a new one @gaede.

vvolkl:
Superseded by #11

PR #11, opened 2019-09-27
[WIP] add SimTrackerHit and CalorimeterHit proposals

gaede:
.. these are the SimTrackerHit and CalorimeterHit  as defined in plcio/LCIO, except with
the two int cellIDs replaced with on 'unsgined long long' as suggested by @vvolkl.
(this PR replaces #6).

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -42,7 +42,6 @@ components:
         int operator[](unsigned i) const { return *( &a + i ) ; }\n
         "
 
-
 datatypes :
 
   edm4hep::MCParticle:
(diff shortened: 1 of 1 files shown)


gaede:
PS: the SimCalorimeterHit is split into the hit itself and  relation to a HitContribution in a different collection (!). This seems necessary as we have a tree structure here: one hit (cell) can have contributions from several MCParticles, which in turn create several steps in the simulation.
I did not manage to map that onto sth.simpler, e.g. using a vector member of a component, as these cannot contain other components or relations ...

vvolkl:
Looks good! This will require AIDASoft/podio#60, i.e. an updated version of podio to compile. Frank, I can make a new cvmfs installation next week if you make a tag of podio, as soon as the last PRs are in of course

gaede:
CI fails due to outdated PODIO.

PR #17, opened 2019-10-14
EDM4hep capitalization convention, Minor text editing

vvolkl:
I propose to stick to the DD4hep spelling convention of writing `EDM4hep` in mixed case, `EDM4HEP` in the context of all caps and `edm4hep` in code for ease of typing. So no `EDM4Hep`, that hopefully means less memorization effort!

BEGINRELEASENOTES
- Fix edm4hep spelling convention

ENDRELEASENOTES


gaede:
This is clearly better...

PR #20, opened 2019-10-25
Skip building podio tests in CI

vvolkl:
Should speed up the build until we have some nightlies

BEGINRELEASENOTES
- Skip building podio tests in CI
ENDRELEASENOTES


vvolkl:
Podio is included in the LCG releases now, so this is not necessarily needed.

vvolkl:
As discussed, it may still be useful to be able to build against the latest PODIO master commit, in order to develop more quickly. In that case this PR is actually still useful.

PR #32, opened 2020-02-11
RDataFrame Helpers

vvolkl:
BEGINRELEASENOTES
- add library + dict with lambdas for use with rdataframe

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,7 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
 
 add_subdirectory(edm4hep)
 add_subdirectory(test)
+add_subdirectory(dataframe)
 
 
 #--- create uninstall target ---------------------------------------------------
--- a/dataframe/CMakeLists.txt
+++ b/dataframe/CMakeLists.txt
@@ -1,0 +1,29 @@
+
+file(GLOB sources *.cc)
+file(GLOB headers *.h)
+
+PODIO_GENERATE_DICTIONARY(libedm4hepRDF edm4hepRDF.h SELECTION selection.xml USES edm4hep)
+
+add_library(edm4hepRDF SHARED ${sources} ${headers} libedm4hepRDF.cxx )
+target_link_libraries(edm4hepRDF edm4hep podio::podio ROOT::Physics)
+
+set_target_properties(edm4hepRDF PROPERTIES
+  PUBLIC_HEADER "${headers}")
+
+install(TARGETS edm4hepRDF
+  # IMPORTANT: Add the library to the "export-set"
+  EXPORT edm4hepTargets
+  RUNTIME DESTINATION "${INSTALL_BIN_DIR}" COMPONENT bin
+  LIBRARY DESTINATION "${INSTALL_LIB_DIR}" COMPONENT shlib
+  PUBLIC_HEADER DESTINATION "${INSTALL_INCLUDE_DIR}/edm4hep"
+    COMPONENT dev)
+
+install(FILES
+  "${PROJECT_BINARY_DIR}/dataframe/libedm4hepRDFDict.rootmap"
+  DESTINATION "${INSTALL_LIB_DIR}" COMPONENT dev)
+
+if (${ROOT_VERSION} GREATER 6)
+  install(FILES
+      "${PROJECT_BINARY_DIR}/dataframe/libedm4hepRDF_rdict.pcm"
+      DESTINATION "${INSTALL_LIB_DIR}" COMPONENT dev)
+endif()
--- a/dataframe/edm4hepRDF.cc
+++ b/dataframe/edm4hepRDF.cc
@@ -1,0 +1,42 @@
+#include "edm4hepRDF.h"
+#include "Math/Vector4D.h"
+
+namespace edm4hep {
+
+std::vector<float> pt (std::vector<MCParticleData> const& in){
+  std::vector<float> result;
+  for (size_t i = 0; i < in.size(); ++i) {
+    result.push_back(std::sqrt(in[i].momentum.x * in[i].momentum.x + in[i].momentum.y * in[i].momentum.y));
+  }
+  return result;
+}
+
+std::vector<float> eta(std::vector<MCParticleData> const& in){
+  std::vector<float> result;
+  ROOT::Math::PxPyPzMVector lv;
+  for (size_t i = 0; i < in.size(); ++i) {
+    lv.SetCoordinates(in[i].momentum.x, in[i].momentum.y, in[i].momentum.z, in[i].mass);
+    result.push_back(lv.Eta());
+  }
+  return result;
+}
+
+std::vector<float> cos_theta(std::vector<MCParticleData> const& in){
+  std::vector<float> result;
+  ROOT::Math::PxPyPzMVector lv;
+  for (size_t i = 0; i < in.size(); ++i) {
+    lv.SetCoordinates(in[i].momentum.x, in[i].momentum.y, in[i].momentum.z, in[i].mass);
+    result.push_back(cos(lv.Theta()));
+  }
+  return result;
+}
+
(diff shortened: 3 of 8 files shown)


vvolkl:
Travis reports  a rather obscure error: 

```
-- Installing: /workspace/install/lib/libedm4hepRDF_rdict.pcm
Test project /workspace/build
    Start 1: write_events
2020/02/11 17:21:35.72 parrot_run[12] <child:1838> fatal: setparrot: fd 5 rfd 4 valid 1 ptr 0x68ba48
2020/02/11 17:21:35.72 parrot_run[12] <child:1838> notice: received signal 15 (Terminated), killing all my children...
2020/02/11 17:21:35.72 parrot_run[12] <child:1838> notice: sending myself 15 (Terminated), goodbye!
/etc/entrypoint.sh: line 59:    12 Terminated              /opt/cctools/bin/parrot_run ${PARROT_OPTIONS} -w $workdir $0 "$@"
travis_time:end:14de2008:start=1581437695218236259,finish=1581438096124451494,duration=400906215235,event=script
[0K[31;1mThe command "docker run -it -v ${PWD}:/workspace -w /workspace  --privileged -e LHCB_ENV_MODE=none -u 0:0 -e USER=root -e GROUP=root -e CMTCONFIG=x86_64-centos7-gcc8-opt  gitlab-registry.cern.ch/lhcb-core/lbdocker/centos7-build:latest bash -c "pwd; ls ;./.edm4hep-ci.d/compile_and_test_with_podio.sh"" exited with 143.[0m
```
This seems to stem from CTest, in particular when I set a dependency between tests (some tests require the output of write_events). I'll just remove the dependency for now and open an issue.

vvolkl:
I'm still puzzled by the error I got previously, but rebasing and cleaning up `CMakeLists.txt` seems to have solved this.

In the lambdas, I switched from the legacy `TLorentzVector` to the recommended `ROOT::Math::LorentzVector`.

vvolkl:
Still removed the file  `dataframe/LinkDef.h` that is replaced by `selection.xml`

dataframe/edm4hepRDF.cc
```diff
@@ -0,0 +1,42 @@
+#include "edm4hepRDF.h"
+#include "Math/Vector4D.h"
+
+namespace edm4hep {
+
+std::vector<float> pt (std::vector<MCParticleData> in){
```
andresailer:
```suggestion
std::vector<float> pt (std::vector<MCParticleData> const& in){
```

dataframe/edm4hepRDF.cc
```diff
@@ -0,0 +1,42 @@
...
+ std::vector<float> result;
+	 for (size_t i = 0; i < in.size(); ++i) {
```
andresailer: The indentation in this file is very inconsitent

dataframe/edm4hepRDF.h
```diff
@@ -0,0 +1,29 @@
...
+#include <cmath>
+#include <vector>
+
+#include "edm4hep/MCParticleData.h"
+#include "edm4hep/SimTrackerHitData.h"
+#include "edm4hep/Vector3f.h"
+#include "edm4hep/Vector3d.h"
+
+
+class TLorentzVector;
```
andresailer: This line isn't needed

dataframe/edm4hepRDF.cc
vvolkl: Agreed, I'll add that to all the methods.

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ execute_process(COMMAND python ${podio_DIR}/../../../python/podio_class_generato
 #--- Declare ROOT dependency ---------------------------------------------------
 list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics)
+find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics ROOTDataFrame)
```
andresailer: I guess this ROOTDataFrame only exists in newer versions of root.
So it might be better to not include this here and thus not force this on everyone

test/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ add_executable(read_events read_events.cc)
 target_include_directories(read_events PUBLIC ${CMAKE_SOURCE_DIR}/edm4hep )
 target_link_libraries(read_events edm4hep podio::podio edm4hepDict )
 add_test(NAME read_events COMMAND read_events)
+set_tests_properties(read_events PROPERTIES DEPENDS write_events)
+
+add_executable(test_rdf test_rdf.cc)
```
andresailer: Instead of requiring the component above we can just check that the library exists
IF(TARGET ROOT::ROOTDataFrame)
  add_executable(....
...
ENDIF()
```

File: CMakeLists.txt (hunk 4 above)
Comment (vvolkl) [reply CMakeLists.txt]: Ok, I'll add a cmake  option for building `dataframe` subdirectory.

File: test/CMakeLists.txt (hunk 5 above)
Comment (vvolkl) [reply test/CMakeLists.txt]: That's much more elegant indeed!

File: dataframe/selection.xml [hunk 6]
```diff
@@ -0,0 +1,11 @@
+<lcgdict>
+    <selection>
+
+            <function name="pt" />
```
andresailer: I don't know anything about this stuff, but shouldn't `cos_theta` be also here?

vvolkl: True, to use  these lambdas from the ROOT interpreter that'll be necessary. Will add it.

dataframe/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+install(TARGETS edm4hepRDF
+  # IMPORTANT: Add the library to the "export-set"
+  EXPORT edm4hepTargets
+  RUNTIME DESTINATION "${INSTALL_BIN_DIR}" COMPONENT bin
+  LIBRARY DESTINATION "${INSTALL_LIB_DIR}" COMPONENT shlib
+  PUBLIC_HEADER DESTINATION "${INSTALL_INCLUDE_DIR}/dataframe"
+    COMPONENT dev)
+
+install(FILES
+  "${PROJECT_BINARY_DIR}/dataframe/libedm4hepRDF.rootmap"
```
andresailer: There is an error in the CI:
```
CMake Error at dataframe/cmake_install.cmake:89 (file):
  file INSTALL cannot find
  "/workspace/build/dataframe/libedm4hepRDF.rootmap".
```

dataframe/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+set_target_properties(edm4hepRDF PROPERTIES
+  PUBLIC_HEADER "${headers}")
+
+install(TARGETS edm4hepRDF
+  # IMPORTANT: Add the library to the "export-set"
+  EXPORT edm4hepTargets
+  RUNTIME DESTINATION "${INSTALL_BIN_DIR}" COMPONENT bin
+  LIBRARY DESTINATION "${INSTALL_LIB_DIR}" COMPONENT shlib
+  PUBLIC_HEADER DESTINATION "${INSTALL_INCLUDE_DIR}/dataframe"
```
andresailer: This should probably also go to include/edm4hep and not to include/dataframe?

dataframe/selection.xml
vvolkl: Also forgot the namespace here that I added only later.

dataframe/CMakeLists.txt
vvolkl: Agreed!

dataframe/CMakeLists.txt
vvolkl: Forgot the `Dict` that becames part of the filename. Fixed.

andresailer:
Ok, looks all good to me now.

PR #34, opened 2020-02-14
CMake: cleanup, use PODIO_GENERATE_DICTIONARY

andresailer:
BEGINRELEASENOTES
- Cmake: remove directory properties, needs AIDASoft/podio#68

ENDRELEASENOTES


andresailer:
Restart build so that the updates from podio are included, build successful now

vvolkl:
I just noticed that the build system won't pick up the local build of podio when using a newer LCG release that includes it, will fix that.

PR #39, opened 2020-02-19
Merging Dicts and Libraries, aligning installation locations with podio

### LINKED ISSUE #37: Align installation locations with podio
The CMake Config etc. of edm4hep is currently installed to `share/cmake`, instead of `lib64/cmake` as for podio.

edm4hep defines installation path as lib instead of lib64

andresailer:
This fixes #37 as far as I am concerned, and other things

BEGINRELEASENOTES
- Fix installation location of EDM4hepVersion.h (EDM4HEP -> edm4hep)
- Change EDM4hep*.cmake install location to `${CMAKE_INSTALL_LIBDIR}/cmake/EDM4HEP`
- Merge edm4hep dictionary library into main edm4hep library
- Merge edm4hepRDF dictionary library into main edm4hepRDF library
- Use CMAKE_INSTALL_*DIR variables defined by GNUInstallDirs

ENDRELEASENOTES


dataframe/CMakeLists.txt
```diff
@@ -2,28 +2,32 @@
...
 
 install(FILES
-  "${PROJECT_BINARY_DIR}/dataframe/libedm4hepRDFDict.rootmap"
-  DESTINATION "${INSTALL_LIB_DIR}" COMPONENT dev)
+  "${PROJECT_BINARY_DIR}/dataframe/edm4hepRDFDict.rootmap"
+  DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT dev)
 
 if (${ROOT_VERSION} GREATER 6)
   install(FILES
       "${PROJECT_BINARY_DIR}/dataframe/libedm4hepRDF_rdict.pcm"
```
vvolkl:
```suggestion
      "${PROJECT_BINARY_DIR}/dataframe/edm4hepRDF_rdict.pcm"
```
I'm guessing the `lib` prefix is removed here too?

vvolkl:
The CI build script seems to end up in the wrong directory and tries to test podio: 

https://travis-ci.com/HSF/EDM4hep/builds/149631798

dataframe/CMakeLists.txt
andresailer: No, the pcm file gets its name from the first argument to PODIO_GENERATE_DICTIONARY, the rootmapfile is set to lib${dictionary}Dict.rootmap. This has to stay to keep avoid to much divergence with the root REFLEX_GENERATE_DICTIONARY

vvolkl: I see! But then it seems  we have to change this line here: https://github.com/HSF/EDM4hep/blob/master/edm4hep/CMakeLists.txt#L42

andresailer: Edited, its of course the other way around

test/hepmc/edm4hep_testhepmc.cc
```diff
@@ -24,12 +24,11 @@ using namespace HepMC;
     edm_particle.setCharge(particleID.charge());
     //  convert momentum
     auto p = hepmcParticle->momentum();
-    //edm_particle.setMomentum( {p.px(), p.py(), p.pz()} );
+    edm_particle.setMomentum( {float(p.px()), float(p.py()), float(p.pz())} );
```
vvolkl: Great, I missed that, thank you!

andresailer:
> The CI build script seems to end up in the wrong directory and tries to test podio:
> 
> https://travis-ci.com/HSF/EDM4hep/builds/149631798

I removed a "cd .." on accident

dataframe/CMakeLists.txt
```diff
@@ -2,28 +2,32 @@
...
 file(GLOB headers *.h)
 
-PODIO_GENERATE_DICTIONARY(libedm4hepRDF edm4hepRDF.h SELECTION selection.xml USES edm4hep)
 
-add_library(edm4hepRDF SHARED ${sources} ${headers} libedm4hepRDF.cxx )
-target_link_libraries(edm4hepRDF edm4hep podio::podio ROOT::Physics)
+add_library(edm4hepRDF SHARED ${sources} ${headers})
+target_link_libraries(edm4hepRDF edm4hep ROOT::Physics)
+PODIO_GENERATE_DICTIONARY(edm4hepRDF ${headers} SELECTION selection.xml USES edm4hepRDF
+  OPTIONS --library edm4hepRDF${CMAKE_SHARED_LIBRARY_SUFFIX}
```
andresailer: @vvolkl infact this was wrong, and makes the install step fail
```suggestion
  OPTIONS --library libedm4hepRDF${CMAKE_SHARED_LIBRARY_SUFFIX}
```

vvolkl: Ah. I was just looking into why the `pcm` files for edm4hep and dataframe were named differently.

dataframe/CMakeLists.txt
andresailer: To be able to merge the dictionary.cxx file into the library we have to tell genreflex that what our library is called, so that name ends up in the rootmap
```
[ libedm4hep.so ]
# List of selected classes
class edm4hep::CaloHitContribution
...
```
Otherwise the GENERATE_DICTIONARY macro uses the libedm4hepDict.so name.
(and changing the macro would diverge from the macro in root, which we maybe can use in the future.)

test/CMakeLists.txt
```diff
@@ -4,19 +4,19 @@ ENDIF()
 
 add_executable(write_events write_events.cc)
 target_include_directories(write_events PUBLIC ${CMAKE_SOURCE_DIR}/edm4hep )
-target_link_libraries(write_events edm4hep podio::podio edm4hepDict )
+target_link_libraries(write_events edm4hep)
```
andresailer: @gaede So I looked at the compile and link commands, and cmake resolves duplicate, e.g., podio::podio dependencies

PR #43, opened 2020-02-24
DDG4EDM4hep Plugin for Use with DDSim

vvolkl:
I tried to extract the commits from the [original PR](https://github.com/AIDASoft/DD4hep/pull/619) by Frank and added some CMake Configuration.
BEGINRELEASENOTES
- add DDG4EDM4hep plugin for use with ddsim

ENDRELEASENOTES

* TODOs: 

- [x] changes in python code from the original PR
- [x] Testing
-  (postponed) CMake Cleanup (possibly in a future PR when updating DD4hep versions)
-  (postponed) Adressing todos in the code and data model changes


gaede:
I actually had overlooked that fact that we also need  changes in the DDSim python code in DD4hep to make this work. Maybe it is then better to leave the plugin in DD4hep and rather implement sth. for the event header information in EDM4hep right away. Will look into this.

andresailer:
We can add the few lines to ddsim independently of the plugin

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,566 @@
...
+	&m_store->get<edm4hep::CaloHitContributionCollection>(colName));
+
+
+    for(unsigned i=0 ; i < nhits ; ++i){
+      auto sch = schc->create() ;
+
+      const Geant4Calorimeter::Hit* hit = coll->hit(i);
+
+      sch.setCellID( hit->cellID );
+      sch.setPosition({float(hit->position.x()/mm),
```
andresailer: Now the "CLHEP::" is dropped from the units, I would prefer if we keep the namespace.

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,566 @@
...
+		       float(hit->position.y()/mm),
+		       float(hit->position.z()/mm)});
+      sch.setEnergy( hit->energyDeposit );
+
+      // now add the individual step contributions
+      for(Geant4HitData::Contributions::const_iterator ci=hit->truth.begin();
+	  ci!=hit->truth.end(); ++ci){
+
+	auto schc = schcc->create();
```
andresailer: The indentation here looks incorrect, tabs again.

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,566 @@
...
+      sth.setPosition({hit->position.x()/CLHEP::mm,
+		       hit->position.y()/CLHEP::mm,
+		       hit->position.z()/CLHEP::mm});
+      sth.setMomentum(edm4hep::Vector3f(hit->momentum.x()/CLHEP::GeV,
+					hit->momentum.y()/CLHEP::GeV,
+					hit->momentum.z()/CLHEP::GeV ));
+
+      auto particleIt = pm->particles().find(trackID);
+      if( ( particleIt != pm->particles().end()) ){
+	// if the original track ID of the particle is not the same as the
```
andresailer: Indentation in this block looks wrong

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,566 @@
...
+  }
+  InstanceCount::decrement(this);
+}
+
+// Callback to store the Geant4 run information
+void Geant4Output2EDM4hep::beginRun(const G4Run* run)  {
+  if ( 0 == m_file && !m_output.empty() )   {
+    G4AutoLock protection_lock(&action_mutex);
+    m_store = new podio::EventStore ;
+    m_file = new podio::ROOTWriter(m_output, m_store);
```
andresailer: Does the ROOTWriter own the m_store after this call?  The m_stire doesn't get cleaned up in the d'tor.

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,566 @@
...
+    m_file->finish();
+    detail::deletePtr(m_file);
+  }
+  InstanceCount::decrement(this);
+}
+
+// Callback to store the Geant4 run information
+void Geant4Output2EDM4hep::beginRun(const G4Run* run)  {
+  if ( 0 == m_file && !m_output.empty() )   {
+    G4AutoLock protection_lock(&action_mutex);
```
andresailer: That lock should go before the `if`?

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
andresailer: This shadows the `schc` from outside the loop. some slightly longer names for sch/schc/schc/schcc should be considered.

vvolkl:
I installed DD4hep in the Key4HEP nightlies in order to test the plugin and speed up the compilation for CI builds. Unfortunately, /cvmfs/sw-nightlies.hsf.org was not mounted in the lhcb docker image we were using so I changed the .travis.yml, more along the lines of DD4hep.

So now this  includes a basic test of ddsim with the edm4hep output plugin.

plugins/DDG4EDM4hep/edm4hep/Geant4Output2EDM4hep.cpp
vvolkl: No it doesn't. I added a `delete` in the destructor.

plugins/DDG4EDM4hep/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  EXPORT DDG4EDM4HEPTargets
+  RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT bin
+  LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
+  PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/edm4hep"
+  COMPONENT dev)
+
+add_test(NAME ddsim_edm4hep COMMAND ddsim --compactFile /cvmfs/sw-nightlies.hsf.org/key4hep/views/latest/x86_64-centos7-gcc8-opt/DDDetectors/compact/SiD.xml -N 2 -G --gun.particle pi+ --outputFile my_edm4hep.root --part.userParticleHandler= )
+set_property(TEST ddsim_edm4hep APPEND PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/edm4hep:$ENV{LD_LIBRARY_PATH}")
+set_property(TEST ddsim_edm4hep APPEND PROPERTY ENVIRONMENT "DD4HEP_LIBRARY_PATH=${CMAKE_BINARY_DIR}/plugins/DDG4EDM4hep/lib/:$ENV{DD4HEP_LIBRARY_PATH}")
+set_property(TEST ddsim_edm4hep APPEND PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/plugins/DDG4EDM4hep/lib/:$ENV{LD_LIBRARY_PATH}")
```
andresailer:
```suggestion
set_tests_properties(ddsim_edm4hep PROPERTIES ENVIRONMENT
LD_LIBRARY_PATH=$<TARGET_FILE_DIR:edm4hep>:$<TARGET_FILE_DIR:DDG4EDM4hep>:$ENV{LD_LIBRARY_PATH}
  DD4HEP_LIBRARY_PATH=$<TARGET_FILE_DIR:DDG4EDM4hep>:$ENV{DD4HEP_LIBRARY_PATH}
)
```

Using the target properties makes this more portable and resilient to changes in the rest of the library, I think.

.travis.yml
```diff
@@ -8,15 +10,53 @@ language: cpp
...
+  - sudo cvmfs_config setup
+  - sudo cvmfs_config probe
+  - sudo mkdir -p /cvmfs/sft.cern.ch
+  - sudo mkdir -p /cvmfs/sft-nightlies.cern.ch
+  - sudo mkdir -p /cvmfs/geant4.cern.ch
+  - sudo mkdir -p /cvmfs/sw-nightlies.hsf.org
+  - ls /cvmfs/sft.cern.ch
+  - ls /cvmfs/sft-nightlies.cern.ch
+  - ls /cvmfs/geant4.cern.ch
+  - ls /cvmfs/clicdp.cern.ch
```
andresailer:
```suggestion
```
The folder for this repo is not created, nor is it used later on.

.travis.yml
```diff
@@ -8,15 +10,53 @@ language: cpp
...
+  - # change wrt dd4hep setup: don't manually mount cvmfs folders
+  - #sudo /etc/init.d/autofs stop
+  - sudo cvmfs_config setup
+  - sudo cvmfs_config probe
+  - sudo mkdir -p /cvmfs/sft.cern.ch
+  - sudo mkdir -p /cvmfs/sft-nightlies.cern.ch
+  - sudo mkdir -p /cvmfs/geant4.cern.ch
+  - sudo mkdir -p /cvmfs/sw-nightlies.hsf.org
+  - ls /cvmfs/sft.cern.ch
+  - ls /cvmfs/sft-nightlies.cern.ch
```
andresailer: This is not part of the CVMFS_REPOS variable given to docker

.travis.yml
```diff
@@ -8,15 +10,53 @@ language: cpp
...
+  - mkdir package
+  - mv !(package) package
+  - shopt -u dotglob
+  - export PKGDIR=${PWD}/package
+
 # command to run tests
-script: 
-  - docker run -it -v ${PWD}:/workspace -w /workspace  --privileged -e LHCB_ENV_MODE=none -u 0:0 -e USER=root -e GROUP=root -e CMTCONFIG=x86_64-centos7-gcc8-opt  gitlab-registry.cern.ch/lhcb-core/lbdocker/centos7-build:latest bash -c "pwd; ls ;./.edm4hep-ci.d/compile_and_test_with_podio.sh"
+script:
+  - docker run -ti --name CI_CONTAINER -v $PKGDIR:/workspace -e COMPILER_TYPE=$COMPILER -e LCG_RELEASE=${LCG_RELEASE} -e STANDARD=${STANDARD} -e COMPILER_VERSION=${COMPILER_VERSION} -e SHARED=${SHARED:-ON} ${CVMFS_REPOS} -d clicdp/cc7-lcg bash
```
andresailer:
```suggestion
  - docker run -ti --name CI_CONTAINER -v $PKGDIR:/workspace -e COMPILER_TYPE=$COMPILER -e LCG_RELEASE=${LCG_RELEASE} -e STANDARD=${STANDARD} -e COMPILER_VERSION=${COMPILER_VERSION} ${CVMFS_REPOS} -d clicdp/cc7-lcg bash 
```

.travis.yml
```diff
@@ -1,5 +1,7 @@
+
+
 sudo: required
```
andresailer: Sudo is deprecated

CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.3)
 
 project(EDM4HEP VERSION 0.1.0)
 
+#--- User-definable options --------------
+
+option(BUILD_DDG4EDM4hep "Build the EDM4hep plugin for DD4hep" ON)
```
andresailer: I think the cmake variables should be all uppercase for reasons of sanity and consistency

plugins/DDG4EDM4hep/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  target_include_directories(DDG4EDM4hep PUBLIC ${DD4hep_INCLUDE_DIRS} ${Geant4_INCLUDE_DIRS})
+  dd4hep_generate_rootmap( DDG4EDM4hep)
+endif()
+
+install(TARGETS DDG4EDM4hep
+  # IMPORTANT: Add the datamodel library to the "export-set"
+  EXPORT DDG4EDM4HEPTargets
+  RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT bin
+  LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
+  PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/edm4hep"
```
andresailer: Does this actually do anything here?

init.sh
```diff
@@ -1,11 +1,13 @@
...
-TOOLSPATH=/cvmfs/fcc.cern.ch/sw/0.8.3/tools/
-OS=`python $TOOLSPATH/hsf_get_platform.py --get os`
+# Disabled as currently there is only one platform supported
+#TOOLSPATH=/cvmfs/fcc.cern.ch/sw/0.8.3/tools/
+#OS=`python $TOOLSPATH/hsf_get_platform.py --get os`
 
-# source FCC externals
-source /cvmfs/fcc.cern.ch/sw/views/releases/externals/96c_LS.0.0/x86_64-$OS-gcc8-opt/setup.sh
+# source LCG releases
+source /cvmfs/sft.cern.ch/lcg/views/LCG_96c_LS/x86_64-centos7-gcc8-opt/setup.sh
```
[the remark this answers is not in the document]
vvolkl: I think it may be a good idea to keep this script working also for the usecase of setting up the environment for compiling dd4hep on, say, lxplus; $LCG_RELEASE and $COMPILER_VERSION are specific to Travis. 
But I changed the script to set those env vars and use them in the path.

vvolkl:
Andre, thanks for the comments! Should now all be addressed.

plugins/DDG4EDM4hep/CMakeLists.txt
vvolkl: No. Removed it.

CMakeLists.txt
vvolkl: Done

.travis.yml
vvolkl: Removed

init.sh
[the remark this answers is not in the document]
andresailer: You can use a default value for the variables
```
${VARIABLE:-DEFAULT}
${LCG_RELEASE:-LCG_96c_LS}
${OS:-centos7}
${COMPILER_VERSION:-gcc8}
```
Or `:=` instead of `:-` if the parameters are used many times, which also assigns the default to the variable, so at the next use the default is used without setting defaults everywhere.

And there is the logic here:
https://github.com/AIDASoft/DD4hep/blob/master/.dd4hep-ci.d/init_x86_64.sh
to switch between slc6 and centos7 
Also some logic to be able to use dev installations from sft-nightlies

.travis.yml
vvolkl: We might still build against the nightlies at one point, but removed it for now.

.edm4hep-ci.d/compile_and_test.sh
```diff
@@ -1,8 +1,23 @@
-
 source init.sh
+
+export CMAKE_PREFIX_PATH=/cvmfs/sw-nightlies.hsf.org/key4hep/packageviews/podio/latest/x86_64-centos7-gcc8-opt:$CMAKE_PREFIX_PATH
```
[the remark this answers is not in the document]
vvolkl: Done

.edm4hep-ci.d/compile_and_test.sh
```diff
@@ -1,8 +1,23 @@
...
+export CMAKE_PREFIX_PATH=/cvmfs/sw-nightlies.hsf.org/key4hep/packageviews/tricktrack/latest/x86_64-centos7-gcc8-opt:$CMAKE_PREFIX_PATH
+export CMAKE_PREFIX_PATH=/cvmfs/sw-nightlies.hsf.org/key4hep/packageviews/DD4hep/latest/x86_64-centos7-gcc8-opt:$CMAKE_PREFIX_PATH
+source /cvmfs/sw-nightlies.hsf.org/key4hep/packageviews/DD4hep/latest/x86_64-centos7-gcc8-opt/bin/thisdd4hep.sh
+
+
+# edm4hep
+mkdir build install
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always " -G Ninja ..  \
```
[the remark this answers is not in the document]
vvolkl: Done

plugins/DDG4EDM4hep/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  COMPONENT dev)
+
+add_test(NAME ddsim_edm4hep COMMAND ddsim --compactFile /cvmfs/sw-nightlies.hsf.org/key4hep/views/latest/x86_64-centos7-gcc8-opt/DDDetectors/compact/SiD.xml -N 2 -G --gun.particle pi+ --outputFile my_edm4hep.root --part.userParticleHandler= )
```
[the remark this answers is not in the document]
vvolkl: Done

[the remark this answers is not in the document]
vvolkl: The detector I use here is not really ideal for a very simple example (takes a while to load and misses some constants, forcing me to use `--part.userParticleHandler=''` maybe we can discuss to include an even simpler test detector with DD4hep?)

vvolkl:
> Nice to have this tested. I have some comments to make this maybe more maintainable, and remove some things not used here.
> 
> Are podio etc build and pushed every day?

Not yet. I'll set up a jenkins job for this, but I wanted to wait until after the discussion tomorrow in case there are any changes to the workflow requested.

init.sh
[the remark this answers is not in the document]
vvolkl: Great suggestion! 

>to switch between slc6 and centos7
Will it be necessary to build on slc6?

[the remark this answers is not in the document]
andresailer: I am still using SLC6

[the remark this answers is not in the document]
vvolkl: Ok. I'd propose to extend the CI to multiple platforms in another PR?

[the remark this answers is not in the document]
andresailer: Yes that is fine.

plugins/DDG4EDM4hep/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+install(TARGETS DDG4EDM4hep
+  EXPORT DDG4EDM4HEPTargets
+  RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT bin
+  LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
+  COMPONENT dev)
+
+add_test(NAME ddsim_edm4hep COMMAND ddsim --compactFile $ENV{DD4hep_DIR}/DDDetectors/compact/SiD.xml -N 2 -G --gun.particle pi+ --outputFile my_edm4hep.root --part.userParticleHandler= )
+set_tests_properties(ddsim_edm4hep PROPERTIES ENVIRONMENT
+  "LD_LIBRARY_PATH=$<TARGET_FILE_DIR:edm4hep>:$<TARGET_FILE_DIR:DDG4EDM4hep>:$ENV{LD_LIBRARY_PATH};DD4HEP_LIBRARY_PATH=$<TARGET_FILE_DIR:DDG4EDM4hep>:$ENV{DD4HEP_LIBRARY_PATH}"
+)
```
andresailer:
```suggestion
  FAIL_REGULAR_EXPRESSION "Error;ERROR;Failed"
)
```

andresailer:
Please rebase

plugins/DDG4EDM4hep/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+include( DD4hep )
+
+dd4hep_configure_output()
+
+#---------------------------  EDM4hep Plugins for new simulation framework  -----------
+# target available with DD4hep > v00-11
+IF(TARGET DD4hep::DDG4)
+  dd4hep_add_plugin(DDG4EDM4hep SOURCES edm4hep/*.cpp)
+  target_link_libraries(DDG4EDM4hep edm4hep DD4hep::DDCore DD4hep::DDG4)
```
andresailer:
```suggestion
  target_link_libraries(DDG4EDM4hep edm4hep podio::podioRootIO DD4hep::DDCore DD4hep::DDG4)
```

.edm4hep-ci.d/compile_and_test.sh
```diff
@@ -1,8 +1,26 @@
-
 source init.sh
+
+export K4PACKAGEVIEW=/cvmfs/sw-nightlies.hsf.org/key4hep/packageviews
+export BINARY_TAG=x86_64-centos7-gcc8-opt
+
+export CMAKE_PREFIX_PATH=${K4PACKAGEVIEW}/podio/latest/${BINARY_TAG}:$CMAKE_PREFIX_PATH
```
andresailer: Try with a key4hep view

vvolkl:
I deliberately introduced an error in the file `/cvmfs/sw-nightlies.hsf.org/key4hep/views/latest/x86_64-centos7-gcc8-opt/include/edm4hep/MCParticle.h` which made the downstream package test fail when I tried it using just the view. This should show if the include order is ok so that the build picks up the local headers first.

vvolkl:
Looks like the key4hep view fails to set `${DD4hep_DIR}`.

vvolkl:
Fixed the key4hep view setup script (I should put this under code review as well).

vvolkl:
@andresailer looks like this picks up all the right  versions/headers now.

PR #44, opened 2020-03-02
Add EventHeader data type

vvolkl:
This is a smaller version of the plcio equivalent, as I'm assuming here that podio will be able to put  parameters like `detectorname` in the metadata tree. Also, removing std::string members means the PODness remains intact.

BEGINRELEASENOTES
- Add EventHeader data type
ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -89,6 +89,7 @@ datatypes :
       - int eventNumber    //event number
       - int runNumber      //run number
       - unsigned long long timeStamp     //time stamp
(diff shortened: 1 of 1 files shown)


gaede:
@vvolkl: the timestamp needs to be `long long` i.e. 64bits. See the LCIO event header:
```
Timestamp is a unique 64 bit integer.<br>
Definition: ns since January 1, 1970, 00:00:00 GMT
```

vvolkl:
True, good catch!

PR #46, opened 2020-03-03
Fix cmake configuration and add test for find_package(EDM4HEP)

### LINKED ISSUE #45: CMake Config Bug
Calling `find_package(EDM4HEP)` with the current configuration results in

```
CMake Error in CMakeLists.txt:

  Imported target "EDM4HEP::edm4hep" includes non-existent path

    "/workspace/install/edm4hep"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not

  provide.

```

Since it is neither complicated nor cpu-expensive, the CI could test for these kinds of issues.

vvolkl:
Fixes #45 and adds a test to catch similar errors in the future.

BEGINRELEASENOTES
- fix cmake target include and add test of edm4hep cmake targets
ENDRELEASENOTES


.edm4hep-ci.d/compile_and_test_with_podio.sh
```diff
@@ -33,3 +33,11 @@ cmake -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_FLAG
       && ninja -k0 \
       && ninja install \
       && ctest --output-on-failure
```
andresailer:
```suggestion
      && ctest --output-on-failure || exit 1
```

edm4hep/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ target_include_directories(edm4hep
   PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/edm4hep>
-  $<INSTALL_INTERFACE:edm4hep>
+  $<INSTALL_INTERFACE:include>
```
andresailer:
```suggestion
  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
```

PR #56, opened 2020-05-18
Standalone Delphes with EDM4HEP output

vvolkl:
This PR disentangles the Delphes program from the ExRootAnalysis TreeWriter, and adds edm4hep output via the podio EventStore instead. The conversion is only a proof of principle now, and still WIP. The branches are created automatically from the `TreeWriter` configuration in the Delphes card.
I started out from `DelphesROOT` and cleared out some non-crucial code and replaced some raw pointers with smart pointers. Ultimately it would be good to make the input/output of the delphes standalone executables more modular (currently quite a bit of code is duplicated in 8 versions of Delphes with different inputs).  
BEGINRELEASENOTES
- Add a converter module for delphes that can output edm4hep files. The module is used in several standalone delphes executables that take different input formats (e.g. stdhep, pythia8, etc.) and **standard delphes cards** and produce edm4hep output root files. It can also be used in conjunction with gaudi. **This is an early release to gather feedback and to iterate on that feedback. Currently not all of the input readers that come with standard delphes are implemented and also the information content of the edm4hep files is restricted to the most important information. More information can be added on demand.** A brief introduction and some examples are included.

ENDRELEASENOTES


tmadlener:
I have rebased the pull request on the current master.
As discussed during todays meeting, I have removed the changes to the `edm4hep::ReconstructedParticle` (#69) and also the dependency on ([podio/95](https://github.com/AIDASoft/podio/pull/95)).

Documentation of how the converter can be configured is still missing.

andresailer:
I think you need to rebase again.

andresailer:
TODO: Change to allow use of un-changed delphes cards

gaede:
@vvolkl and @tmadlener: please add a few lines to the Release notes

tmadlener:
Hi @vvolkl,
Looking at your last commit I see that you introduce one global `ParticleIDCollection`. Would it make sense to have one collection per output collection? I am mainly thinking about the tooling support, like some utility functionality that facilitates accessing the information stored in the `ParticleID`s. With different collection we could put, e.g. the meaning of the different indices, into the metadata of each collection. I am not entirely sure whether we need separate collections for that or whether we could still do that with one collection and slightly more elaborate metadata. 

For now I think this can go in as it is, but we may need to revisit this in a future iteration. The question is mainly: Where do we want to do the bookkeeping.

This also needs to be rebased onto master again if we do not want to have a merge commit in the history.

vvolkl:
Hi Thomas,
Yes, I thought about this, but at the moment I don't see the need, and with one global collection it is simpler. But we can indeed revisit this in the future.

For me the rebase and merge button still works - I'll do one more round of testing, then I'll merge.

PR #58, opened 2020-05-18
Rename cluster vectormembers in plural form

vvolkl:
Plural forms for the variable names make it easier to recognize they refer to a vector.

BEGINRELEASENOTES
- Rename cluster vectormembers with a plural form

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -269,8 +269,8 @@ datatypes :
       - float                phi             //intrinsic direction of cluster at position - Phi. Not to be confused with direction cluster is seen from IP. 
       - edm4hep::Vector3f    directionError  //covariance matrix of the direction (3 Parameters)
     VectorMembers:
-      - float   shape                //shape parameters - check/set collection parameter ClusterShapeParameters for size and names of parameters.
-      - float   weight               //weight of a particular cluster
+      - float   shapeParameters      //shape parameters - check/set collection parameter ClusterShapeParameters for size and names of parameters.
+      - float   weights              //weight of a particular cluster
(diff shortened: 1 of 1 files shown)


vvolkl:
@gaede While changing this I realized I'm not entirely sure how these fields should be used. The `weights` array refers to the `clusters` relation, which can be used to combine clusters to a bigger one? And I think we don't have collection parameters yet to explain a bit more how shapeParameters should be used, this is another example where they would be useful.

gaede:
I have to admit, I don't know where the `weights` field comes from - it is not in LCIO but in pLCIO.
To my knowledge, we are not really using `Clusters` made from `Clusters` in iLCSoft, so far.
Maybe we should drop the weights for now ?
I agree that meta data would be extremely useful - working on it...

edm4hep.yaml
```diff
@@ -269,8 +269,8 @@ datatypes :
       - float                phi             //intrinsic direction of cluster at position - Phi. Not to be confused with direction cluster is seen from IP. 
       - edm4hep::Vector3f    directionError  //covariance matrix of the direction (3 Parameters)
     VectorMembers:
-      - float   shape                //shape parameters - check/set collection parameter ClusterShapeParameters for size and names of parameters.
-      - float   weight               //weight of a particular cluster
+      - float   shapeParameters                //shape parameters - check/set collection parameter ClusterShapeParameters for size and names of parameters.
```
andresailer, line 272:
```suggestion
      - float   shapeParameters      //shape parameters - check/set collection parameter ClusterShapeParameters for size and names of parameters.
```
To align the comments, I  could have missed the correct  number of spaces.

PR #66, opened 2020-06-09
Introduce CMake function to enforce proper linking of Dicts

tmadlener:
BEGINRELEASENOTES
- Introduce convenience cmake function for ensuring that dictionaries are properly linked on Ubuntu

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,28 @@ endif()
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -DDROP_CGAL ${CPP_STANDARD_FLAGS} -Wall -Wextra -Wpedantic -Wno-unused-variable -Wno-unused-parameter")
 
 
+#=====================================================================
+# check if the linker supports as needed or not
+#   (needed for ROOT dict, specifically on Ubuntu)
+set(STORE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+include(CheckCCompilerFlag)
+set(CMAKE_REQUIRED_FLAGS "-Wl,--no-as-needed")
+check_c_compiler_flag("" TEST_NO_AS_NEEDED)
+set(CMAKE_REQUIRED_FLAGS "-Wl,--as-needed")
+check_c_compiler_flag("" TEST_AS_NEEDED)
+set(CMAKE_REQUIRED_FLAGS ${STORE_REQUIRED_FLAGS})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -60,6 +60,28 @@ endif()
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -DDROP_CGAL ${CPP_STANDARD_FLAGS} -Wall -Wextra -Wpedantic -Wno-unused-variable -Wno-unused-parameter")
 
 
+#=====================================================================
+# check if the linker supports as needed or not
+#   (needed for ROOT dict, specifically on Ubuntu)
+set(STORE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+include(CheckCCompilerFlag)
+set(CMAKE_REQUIRED_FLAGS "-Wl,--no-as-needed")
+check_c_compiler_flag("" TEST_NO_AS_NEEDED)
```
andresailer, line 69: the variables are added to the internal cache (https://cmake.org/cmake/help/latest/module/CheckCCompilerFlag.html) So it should be perfectly fine to include these checks inside the function body.

CMakeLists.txt
```diff
@@ -60,6 +60,28 @@ endif()
...
+set(CMAKE_REQUIRED_FLAGS "-Wl,--as-needed")
+check_c_compiler_flag("" TEST_AS_NEEDED)
+set(CMAKE_REQUIRED_FLAGS ${STORE_REQUIRED_FLAGS})
+
+function(target_link_libraries_force_dict target dict)
+  if(TEST_NO_AS_NEEDED AND TEST_AS_NEEDED)
+    target_link_libraries(${target}
```
andresailer, line 76: I am not a fan of hiding these basic function calls. 
Also it might be more flexible to use the function just to add the push/pop flags around the argument and return the decorated result, this way one could still decide on the order of linked libraries without having the dict first

tmadlener: Ah yes, that makes sense. So basically something like a `force_link_dict(dict)` that decorates the dictionary, such that in the end we could do:
```
force_link_dict(edm4hepDict)
target_link_library(read_events edm4hep edm4hepDict podio::podioRootIO)
```
which would leave the `target_link_library` call untouched and would also allow to force link more than one dictionary per call?

tmadlener:
This is more cleanly addressed by #68

PR #68, opened 2020-06-16
Fix rootmap

andresailer:
Analogous to https://github.com/AIDASoft/podio/pull/99

BEGINRELEASENOTES
- The rootmap file should point to the dictionary library, and not the base library

- Tests: Fill the LD_LIBRARY_PATH environment variable with the location of the rootmap file

ENDRELEASENOTES

### DIFF
--- a/edm4hep/CMakeLists.txt
+++ b/edm4hep/CMakeLists.txt
@@ -29,7 +29,7 @@ target_include_directories(edm4hepDict
 )
 
 PODIO_GENERATE_DICTIONARY(edm4hepDict ${headers} SELECTION src/selection.xml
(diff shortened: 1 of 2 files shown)


tmadlener:
This solves the original problem on Ubuntu 18.04 and 20.04 for me.

vvolkl:
Very good to have this issue fixed! As far as I can tell, setting the LD_LIBRARY_PATH to include the rootmap files is mandatory, not even installing in a directory in  /usr/local or similar will allow writing edm4hep root files. Or is there another possibilitiy I overlooked? If not, maybe it is good to add a line to the README to document this.

PR #69, opened 2020-06-18
Add more description concerning the internal 4-momentum state of ReconstructedParticle

tmadlener:
BEGINRELEASENOTES
- Add additional comment to `mass`, `energy` and `momentum` members of `ReconstructedParticle` description to clearly state that the 4-momentum state is not automatically kept consistent internally. Since there are too many different use-cases, a general solution that always works was deemed hard to achieve and it is, thus, the users responsibility to keep this state consistent if it is necessary.

ENDRELEASENOTES

It is still necessary to set the mass after the momentum has been set to get everything really consistent, but without a 4-vector type, I think that is hard to achieve.

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -351,11 +351,11 @@ datatypes :
     Author : "F.Gaede, DESY"
     Members:
       - int                    type           //type of reconstructed particle. Check/set collection parameters ReconstructedParticleTypeNames and ReconstructedParticleTypeValues.
-      - float                  energy         // [GeV] energy of the reconstructed particle.
-      - edm4hep::Vector3f      momentum       // [GeV] particle momentum
+      - float                  energy         // [GeV] energy of the reconstructed particle. Four momentum state is not kept consistent internally.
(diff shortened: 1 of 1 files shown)


gaede:
Hi @tmadlener , the getMass() is in principle fine, if we decide to drop the mass member. The setMass(), implicitly setting the Energy we should not have, as there is a setEnergy() method.
In  case of neutral particles, the typical reconstruction flow would be to take set the meassured Cluster energy - and then compute the 3-momentum from the cluster position, energy and a mass hypothesis. One could think about having a  `setMassAndEnergy()` and `setMassAndMomentum()`  ensuring consistency ...

tmadlener:
Ah yes, that is a good point. I introduced `setMass()` mainly for having some sort of symmetry to the `MCParticle`, where the mass is the member variable. But I agree, that it would be much better and cleaner to only have a method that either sets the complete 4-vector (like `setMassAndMomentum()` or the ones that modify the actual members.

tmadlener:
Following the discussion we had today, I have reverted all changes, except for some additional documentation in the comments.

PR #70, opened 2020-07-03
replace const cast with registerCollection

vvolkl:


tmadlener:
CI problems seem to be unrelated and it works locally. Additionally #56 is still under review, so this will not go directly to `master` in any case.

PR #72, opened 2020-07-15
Fix the declaration of extracode in TrackState

### LINKED ISSUE #71: TrackState static members are not generated
The const static members of the `TrackState` component are not generated. According to the definition:

```yaml
TrackState:
  # some more declarations
  ExtraCode:
    const_declaration: "
      static const int AtOther = 0 ; // any location other than the ones defined below\n   
      static const int AtIP = 1 ;\n
      static const int AtFirstHit = 2 ;\n                   
      static const int AtLastHit = 3 ;\n                  
      static const int AtCalorimeter = 4 ;\n                
      static const int AtVertex = 5 ;\n     
      static const int LastLocation = AtVertex  ;\n
      "
```

should be present, but they are not generated. E.g. `grep AtOther edm4hep/edm4hep/*.h` turns up empty.

The reason is that `const_declaration` is ignored by the `podio` class generator for components. See AIDASoft/podio#115

tmadlener:
BEGINRELEASENOTES
- Change `const_declaration` to `declaration` in `ExtraCode` to have static members of `TrackState` included in generated code (See #71 )

ENDRELEASENOTES

Fixes #71

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -58,7 +58,7 @@ components:
       referencePoint : edm4hep::Vector3f # Reference point of the track parameters, e.g. the origin at the IP, or the position  of the first/last hits or the entry point into the calorimeter.
       covMatrix      : std::array<float,15> # upper triangular covariance matrix of the track parameters.  the order of parameters is:   d0, phi, omega, z0, tan(lambda). 
       ExtraCode :
-        const_declaration: "
+        declaration: "
         static const int AtOther = 0 ; // any location other than the ones defined below\n   
         static const int AtIP = 1 ;\n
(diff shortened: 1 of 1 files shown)


vvolkl:
This is a workaround for https://github.com/AIDASoft/podio/issues/115 right? Fine for me, but I think it should be addressed in podio as well.

tmadlener:
Yes, but I think this is in fact the correct way to declare this. Since there are no `Const` components, `const_declarations` are a bit meaningless in the `ExtraCode` of components.

This should in any case work even after AIDASoft/podio#115 has been addressed.

tmadlener:
Now incorporated in #75

PR #79, opened 2020-09-03
Fix delphes EDM4Hep plugin

gaede:
BEGINRELEASENOTES
- fix delphes EDM4Hep plugin
     - set energy for clusters and reco particles in DelphesEDM4HepConverter
     - add example higgs_recoil_plots.C  (same as in delphes2lcio)
     - minor fix for building DelphesPythia8 on MacOS
     - make plugin compatible w/ Pythia8.3

ENDRELEASENOTES

### DIFF
--- a/plugins/delphes/DelphesPythia8Reader.h
+++ b/plugins/delphes/DelphesPythia8Reader.h
@@ -209,12 +209,14 @@ class DelphesPythia8Reader: public DelphesInputReader {
     pythia = new Pythia8::Pythia;
 
     // jet matching
+#if PYTHIA_VERSION_INTEGER < 8300
     matching = combined->getHook(*pythia);
     if(!matching)
     {
-      throw std::runtime_error("can't do matching");
+      throw runtime_error("can't do matching");
     }
     pythia->setUserHooksPtr(matching);
+#endif
 
     if(pythia == NULL)
     {
--- a/plugins/delphes/DelphesPythia8Reader.h
+++ b/plugins/delphes/DelphesPythia8Reader.h
@@ -213,7 +213,7 @@ class DelphesPythia8Reader: public DelphesInputReader {
     matching = combined->getHook(*pythia);
     if(!matching)
     {
-      throw runtime_error("can't do matching");
+      throw std::runtime_error("can't do matching");
     }
     pythia->setUserHooksPtr(matching);
 #endif
--- a/plugins/delphes/examples/higgs_recoil_plots.C
+++ b/plugins/delphes/examples/higgs_recoil_plots.C
@@ -12,14 +12,14 @@
 
 /*
 put this into your .rootlogon.C file
-
+and make sure the (DY)LD_LIBRARY_PATH contains the podio and edm4hep installations
 {
  gInterpreter->AddIncludePath("$PODIO/include");
- gSystem->Load("$PODIO/lib/libpodio.so");
- gSystem->Load("$PODIO/lib/libpodioDict.so");
- gSystem->Load("$PODIO/lib/libpodioRootIO.so");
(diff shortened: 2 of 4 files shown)


plugins/delphes/examples/higgs_recoil_plots.C
```diff
@@ -0,0 +1,133 @@
...
+/*
+put this into your .rootlogon.C file
+
+{
+ gInterpreter->AddIncludePath("$PODIO/include");
+ gSystem->Load("$PODIO/lib/libpodio.so");
+ gSystem->Load("$PODIO/lib/libpodioDict.so");
+ gSystem->Load("$PODIO/lib/libpodioRootIO.so");
+ gSystem->Load("$EDM4HEP/lib/libedm4hep.dylib");
+ gSystem->Load("$EDM4HEP/lib/libedm4hepDict.dylib");
```
tmadlener, line 22: Are these endings mac specific for these libs?
Could we add these lines to the README?

plugins/delphes/examples/higgs_recoil_plots.C
```diff
@@ -0,0 +1,133 @@
...
+  return TLorentzVector( p.getMomentum()[0],p.getMomentum()[1], p.getMomentum()[2],p.getEnergy());
+}
+
+/** Example script for creating invariant mass plots from a ee -> HZ-> X mu mu sample
+ *  
+ *  A suitable input file for the ILC can be found here: 
+ *  http://osggridftp02.slac.stanford.edu:8080/sdf/group/lcddata/ilc/prod/ilc/mc-dbd/generated/250-TDR_ws/higgs/E250-TDR_ws.Pe2e2h.Gwhizard-1_95.eR.pL.I106480.001.stdhep
+ *  
+ *  Use this together with DelphesSTDHEP_EDM4HEP and the delphes card from https://github.com/iLCSoft/ILCDelphes.git.
+ */
```
tmadlener, line 47: I would also add this information to the README

plugins/delphes/examples/higgs_recoil_plots.C
vvolkl: I think you can remove the endings, root will try to append `[.so | .dll | .dylib | .sl | .dl | .a] ` if nothing is given. Also, the absolute path is not necessary - the libraries just need to be installed to `LD_LIBRARY_PATH`. 

I'm not sure we can rely on `$PODIO` and `$EDM4HEP` being set  - they are definitely not set in the spack build.

tmadlener: Ah I didn't know that root tries to resolve these automatically. In any case the [README](https://github.com/key4hep/EDM4hep/blob/master/plugins/delphes/examples/README.md) of the delphes examples mentions the requirements for the `LD_LIBRARY_PATH`.

Maybe loading all the libraries could be placed directly in this example as it is done for the other examples?

plugins/delphes/DelphesPythia8Reader.h
```diff
@@ -209,12 +209,14 @@ class DelphesPythia8Reader: public DelphesInputReader {
     pythia = new Pythia8::Pythia;
 
     // jet matching
+#if PYTHIA_VERSION_INTEGER < 8300
     matching = combined->getHook(*pythia);
     if(!matching)
     {
-      throw std::runtime_error("can't do matching");
+      throw runtime_error("can't do matching");
```
vvolkl, line 216: This line trips up the CI. I think the namespace shouldn't be removed.
```suggestion
      throw std::runtime_error("can't do matching");
```

gaede: yep - copy paste error...

PR #82, opened 2020-10-05
[delphes] convert particle status

vvolkl:
BEGINRELEASENOTES
- save the delphes particle status as edm4hep::MCparticle.GeneratorStatus

ENDRELEASENOTES

### DIFF
--- a/plugins/delphes/DelphesEDM4HepConverter.cc
+++ b/plugins/delphes/DelphesEDM4HepConverter.cc
@@ -139,8 +139,7 @@ void DelphesEDM4HepConverter::processParticles(const TObjArray* delphesCollectio
         (float) delphesCand->Position.Y(),
         (float) delphesCand->Position.Z()});
     cand.setPDG(delphesCand->PID); // delphes uses whatever hepevt.idhep provides
-    // TODO: - status
-    // TODO: - ...
+    cand.setGeneratorStatus(delphesCand->Status);
 
     if (const auto [it, inserted] = m_genParticleIds.emplace(delphesCand->GetUniqueID(), cand); !inserted) {
       std::cerr << "**** WARNING: UniqueID " << delphesCand->GetUniqueID() << " is already used by MCParticle with id: " << it->second.id() << std::endl;


tmadlener:
This looks good to me. Should we add a link to delphes to the README in order to make it easier on how to interpret this value? (Or is that "common knowledge"?)

clementhelsens:
> This looks good to me. Should we add a link to delphes to the README in order to make it easier on how to interpret this value? (Or is that "common knowledge"?)

It does not comes from Delphes only.
For example if you run pythia, you will have statuses according to what pythia8 does:
http://home.thep.lu.se/~torbjorn/pythia81html/ParticleProperties.html

tmadlener:
Ah, I see. I am not sure if I understand the Delphes documentation on the `Candidate::Status`. Is it only filled from the generator/input and not changed by Delphes after that? That is what I understand, but I am not entirely sure whether that is actually the case. But if it is the case, than we could simply say that we use the same value that the generator provides.

Or maybe even less specific: We just use what Delphes gives us and it is than users that know how to use it with Delphes output would also know how to use it in the edm4hep case?

vvolkl:
In the case of DelphesPythia8, the status of the candidate is set like this: https://github.com/key4hep/EDM4hep/blob/master/plugins/delphes/DelphesPythia8Reader.h#L92
I'm not sure if the `Status` of DelphesCandidate is used otherwise, but for this usecase this seems like the right thing.

vvolkl:
CI failing because I cleaned up the setup scripts on cvmfs - now points to the one that is also mentioned in the docs.

tmadlener:
> In the case of DelphesPythia8, the status of the candidate is set like this: https://github.com/key4hep/EDM4hep/blob/master/plugins/delphes/DelphesPythia8Reader.h#L92
> I'm not sure if the `Status` of DelphesCandidate is used otherwise, but for this usecase this seems like the right thing.

Yes, I agree.

tmadlener (approved):
Is the CI still broken, or are we just seeing an old status?

PR #83, opened 2020-10-05
[delphes] Use the branch name instead of the delphes class for GenParticle

tmadlener:
Otherwise we would select on the delphes class, which could lead to the
problem that we could accidentally select more branches than we had
anticipated. There would be not possibility to only select a subset of
all branches that correspond to the specified delphes class. Now we can
instead select the different branches by their name, which is also
consistent with how the other configuration parameters work.

BEGINRELEASENOTES
- Make delphes output configuration more consistent, by using branch names also for the generated particle collections.

ENDRELEASENOTES


vvolkl (approved):

vvolkl:
Closing and reopening to trigger CI

PR #85, opened 2020-10-13
[ci] add lib64 for local podio to ld_library_path

vvolkl:
BEGINRELEASENOTES
- [ci] fix the library path for builds with a local podio

ENDRELEASENOTES


andresailer:
Looking through the libraries,
This is the only one going to /lib 
https://travis-ci.com/github/key4hep/EDM4hep/jobs/385056062#L769-L771
Maybe fix the CMAKE for the GnuInstallDirs
https://github.com/key4hep/EDM4hep/blob/master/plugins/delphes/CMakeLists.txt#L11

vvolkl:
@andresailer yes, installing everything in the same directory is a good suggestion, done. I kept `lib` in the CI setup because that is plattform dependent, I guess? For the record, ddsim and other applications using the dictionaries also need to have either CPATH or ROOT_INCLUDE_PATH set to pick up changed headers.

PR #90, opened 2020-11-03
Clear internal maps before the referenced objects are clered by the EventStore

### LINKED ISSUE #89: Segmentation violation running DelphesPythia8_EDM4HEP
While running large number of jobs on HTcondor, I noticed that a large fraction of them where not completed, the number of events was smaller than what it is supposed to be. Looking at the log files one can see a segmentation violation occurring after a random number of events are generated. See at the end the full log.

To reproduce this, just do the following:

source /cvmfs/fcc.cern.ch/sw/latest/setup.sh
DelphesPythia8_EDM4HEP delphes_card_IDEAtrkCov.tcl edm4hep_output_config.tcl ee_Z_bbbar.cmd out.root

For your convenience all the input files can be found at:
/afs/cern.ch/user/h/helsens/public/EDM4HEP

It will run and eventually fail after a more or less large number of events.

 *** Break *** segmentation violation
```
===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007f903bd7a46c in waitpid () from /lib64/libc.so.6
#1  0x00007f903bcf7f62 in do_system () from /lib64/libc.so.6
#2  0x00007f903ce15487 in Exec (shellcmd=<optimized out>, this=0x197f630) at /tmp/vavolkl/spack-stage/spack-stage-root-6.20.04-rdccmbcnwuuiotc7behireo7a7ohc5ng/spack-src/core/unix/src/TUnixSystem.cxx:2107
#3  TUnixSystem::StackTrace() () at /tmp/vavolkl/spack-stage/spack-stage-root-6.20.04-rdccmbcnwuuiotc7behireo7a7ohc5ng/spack-src/core/unix/src/TUnixSystem.cxx:2397
#4  0x00007f903ce175d4 in TUnixSystem::DispatchSignals (this=0x197f630, sig=kSigSegmentationViolation) at /tmp/vavolkl/spack-stage/spack-stage-root-6.20.04-rdccmbcnwuuiotc7behireo7a7ohc5ng/spack-src/core/unix/src/TUnixSystem.cxx:3628
#5  <signal handler called>
#6  0x00007f9040196999 in edm4hep::ConstMCParticle::~ConstMCParticle() () from /cvmfs/sw.hsf.org/spackages/linux-centos7-broadwell/gcc-8.3.0/edm4hep-0.2.1-a456et23b4wtglltnr3atttpddxg6uge/lib64/libedm4hep.so
#7  0x00007f9041524ebd in DelphesEDM4HepConverter::process(Delphes*) () from /cvmfs/sw.hsf.org/spackages/linux-centos7-broadwell/gcc-8.3.0/edm4hep-0.2.1-a456et23b4wtglltnr3atttpddxg6uge/lib64/libDelphesEDM4HepConverter.so
#8  0x000000000041bc12 in doit(int, char**, DelphesInputReader&) ()
#9  0x0000000000418417 in main ()
===========================================================


The lines below might hint at the cause of the crash.
You may get help by asking at the ROOT forum http://root.cern.ch/forum
Only if you are really convinced it is a bug in ROOT then please submit a
report at http://root.cern.ch/bugs Please post the ENTIRE stack trace
from above as an attachment in addition to anything else
that might help us fixing this issue.
===========================================================
#6  0x00007f9040196999 in edm4hep::ConstMCParticle::~ConstMCParticle() () from /cvmfs/sw.hsf.org/spackages/linux-centos7-broadwell/gcc-8.3.0/edm4hep-0.2.1-a456et23b4wtglltnr3atttpddxg6uge/lib64/libedm4hep.so
#7  0x00007f9041524ebd in DelphesEDM4HepConverter::process(Delphes*) () from /cvmfs/sw.hsf.org/spackages/linux-centos7-broadwell/gcc-8.3.0/edm4hep-0.2.1-a456et23b4wtglltnr3atttpddxg6uge/lib64/libDelphesEDM4HepConverter.so
#8  0x000000000041bc12 in doit(int, char**, DelphesInputReader&) ()
#9  0x0000000000418417 in main ()
===========================================================
```

tmadlener:
Avoid running into a double free, because the EventStore::clearCollections that might be called outside will free the objects wo which we still hold pointers and would free again when we deconstruct the objects that we hold int he maps.

BEGINRELEASENOTES
- Fix double free in Delphes EDM4Hep converter that lead to spurious segmentation violations (#89)

ENDRELEASENOTES

Fixes #89

### DIFF
--- a/plugins/delphes/DelphesEDM4HepConverter.cc
+++ b/plugins/delphes/DelphesEDM4HepConverter.cc
@@ -103,8 +103,6 @@ void DelphesEDM4HepConverter::process(Delphes *modularDelphes) {
   for (auto& coll : m_collections) {
    coll.second->clear();
   }
-  m_genParticleIds.clear();
-  m_recoParticleGenIds.clear();
 
   for (const auto& branch : m_branches) {
     const auto* delphesCollection = modularDelphes->ImportArray(branch.input.c_str());
@@ -119,6 +117,11 @@ void DelphesEDM4HepConverter::process(Delphes *modularDelphes) {
       (this->*processFuncIt->second)(delphesCollection, branch.name);
     }
   }
+
+  // Clear the internal maps that hold references to entites that have been put
+  // into maps here for internal use only (see #89)
+  m_genParticleIds.clear();
+  m_recoParticleGenIds.clear();
 }
 
 void DelphesEDM4HepConverter::processParticles(const TObjArray* delphesCollection, std::string_view const branch)


tmadlener:
I have run some more tests and this seems to indeed fix the problem.

@clementhelsens can you check whether this does also solve #89 for you?

clementhelsens:
Yes, I will check later today with jobs that usually had a large failure rate.

clementhelsens:
Just checked with 100 jobs that failed basically 100% of the time before and just got 100% success now!
will merge, thanks again

PR #93, opened 2020-11-08
Remove call to Delphes treeWriter

clementhelsens:
BEGINRELEASENOTES
This patch is to avoid a segmentation fault when calling the Delphes treeWriter only occurring when reading a LHE file from Pythia8
ENDRELEASENOTES


vvolkl:
Yes, this looks like it should be removed here, good catch

PR #94, opened 2020-11-09
Remove Delphes Plugin (migrated to k4simdelphes)

vvolkl:
BEGINRELEASENOTES
- Remove Delphes plugin (which is migrated to its own repository key4hep/k4simdelphes)

ENDRELEASENOTES


vvolkl:
Adding the evtgen reader without declaring the dependency in spack breaks the nightlies. Merging this to fix them.

Issue #42, opened 2020-02-24
Build with LCG nightlies not working


vvolkl:
Trying to build with `dev3` fails with the cmake error:

```
~/repo/EDM4hep/build (ddg4_plugin *)$ source /cvmfs/sft.cern.ch/lcg/views/dev3/Thu/x86_64-centos7-gcc8-opt/setup.sh
~/repo/EDM4hep/build (ddg4_plugin *)$ cmake ..
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/gcc
-- Check for working C compiler: /cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++
-- Check for working CXX compiler: /cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /cvmfs/sft-nightlies.cern.ch/lcg/views/dev3/Thu/x86_64-centos7-gcc8-opt/cmake/ROOTConfig-targets.cmake:1276 (message):
  The imported target "ROOT::cppyy_backend" references the file

     "/cvmfs/sft-nightlies.cern.ch/lcg/views/dev3/Thu/x86_64-centos7-gcc8-opt/lib/python2.7/site-packages/libcppyy_backend.so"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/cvmfs/sft-nightlies.cern.ch/lcg/views/dev3/Thu/x86_64-centos7-gcc8-opt/cmake/ROOTConfig-targets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
[...]

PR #96, opened 2020-11-11
use spack for local build of podio

vvolkl:
BEGINRELEASENOTES
- Use spack dev-build in the CI for local podio
- ~fix dev3 by setting root_include_path~

ENDRELEASENOTES

### DIFF
--- a/.edm4hep-ci.d/compile_and_test_with_spack.sh
+++ b/.edm4hep-ci.d/compile_and_test_with_spack.sh
@@ -1,0 +1,14 @@
+mkdir build
+cd build
+git clone --depth 1 https://github.com/key4hep/spack
+git clone --depth 1 https://github.com/key4hep/key4hep-spack
+cp key4hep-spack/config/packages.yaml spack/etc/spack
+cp key4hep-spack/config/cvmfs_build/upstreams.yaml spack/etc/spack
+source spack/share/spack/setup-env.sh
+spack repo add --scope site key4hep-spack
+# setup compiler for spack to find
+source /cvmfs/sw.hsf.org/key4hep/setup.sh
+spack compiler find --scope site
+cd ..
+spack dev-build --test=all edm4hep@master~ddg4_edm4hep_plugin ^podio@master
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,16 +9,15 @@ language: cpp
 
 env:
   matrix:
-    - SETUPSCRIPT=/cvmfs/sw.hsf.org/key4hep/setup.sh; STANDARD=17; BUILD_DDG4EDM4HEP=ON; PODIO_FROM=STACK; STACK=KEY4HEPNIGHTLY;
-    - SETUPSCRIPT=/cvmfs/sw.hsf.org/key4hep/setup.sh; STANDARD=17; BUILD_DDG4EDM4HEP=ON; PODIO_FROM=BUILD_HEAD_ONTHEFLY; STACK=KEY4HEPNIGHTLY;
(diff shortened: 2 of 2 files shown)


vvolkl:
Fixes #42

.edm4hep-ci.d/compile_and_test_with_spack.sh
```diff
@@ -0,0 +1,13 @@
+source .edm4hep-ci.d/setup_spack_gcc.sh
```
tmadlener, line 1: Can we do this on the fly somehow, so that it doesn't depend on hardcoded paths in the setup script? I see you have tried to do it via `spack` loads, but opted for this setup script, why is that?

tmadlener (commented):
Is there a way to generate the `setup_spack_gcc.sh` script on the fly?

.edm4hep-ci.d/compile_and_test_with_spack.sh
vvolkl: Yes, I'm not sure what happens there, I think because maybe the docker image has no system g++, spack does not load gcc8.3 (which was compiled with the centos7 gcc). I can move this setup script to cvmfs.

tmadlener: Ah yes, it kind of makes sense, that spack will not be able to locate the spack installed `gcc`, if it doesn't know with which compiler it has been built originally. We could potentially try to trick it into assuming that `gcc@4.8.5` is here by just filling the `.spack/linux/compilers.yaml` accordingly, but that is also brittle if the central spack installation changes.

In any case such a `setup_spack_gcc.sh` script might be nice to have in cvmfs, e.g. when one just needs a compiler and not the whole stack.

andresailer:
The problem is that the listcomponents used to create the components file, is that from Gaudi, not from DD4hep
```
ls -l /cvmfs/sft.cern.ch/lcg/views/dev3/latest/x86_64-centos7-gcc8-opt/bin/listcomponents
lrwxrwxrwx 1 cvmfs cvmfs 107 Nov 24 05:12 /cvmfs/sft.cern.ch/lcg/views/dev3/latest/x86_64-centos7-gcc8-opt/bin/listcomponents -> /cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Tue/Gaudi/master/x86_64-centos7-gcc8-opt/bin/listcomponents
```

The components file is empty, and the dd4hep plugin manager doesn't find the plugin

vvolkl:
Indeed, in LCG98 listcomponents links to the DD4hep provided one. To be fixed in dev3? (Is it actually a good idea for that executable to have the same name if they are not compatible?)

andresailer:
I rename the one in dd4hep

vvolkl:
Also I'm somewhat surprised that `listcomponents` is taken from PATH. When the cmake config of dd4hep should know where its listcomponents is located, so another solution would be to invoke it with the full path.

andresailer:
It is not take from PATH, but the dd4hepConfig-targets-release.cmake is picking up the file relative to its location in the view (relocatable!), not via its actual location. At least at the moment. As people shouldn't manually call listcomponents, but only via dd4hep_add_plugin, where the target is being used. Which is why renaming the executable on disk should be transparent.

andresailer:
Needs rebase, if still valid

vvolkl:
Rebased, but left out the dev3 fixes which need to be checked again, and probably deserve their own PR. Maybe I should also take the opportunity to switch the CI to github actions.

vvolkl:
Closing this because of https://github.com/key4hep/key4hep-spack/issues/234

PR #101, opened 2020-12-03
Remove DDG4 plugin (moves to dd4hep)

vvolkl:
In preparation for the dd4hep release, which shifts the dependencies, it will be good to have a version of edm4hep without dd4hep dependencies (even optional). So we should  make a new tag, incrementing the minor release number after merging this.
 
BEGINRELEASENOTES
- Remove DDG4 plugin (moves to upstream dd4hep, https://github.com/AIDASoft/DD4hep/pull/740)

ENDRELEASENOTES


andresailer:
Related DD4hep PR: https://github.com/AIDASoft/DD4hep/pull/740

andresailer (approved):

PR #117, opened 2021-06-29
WIP: An example to show the User Defined data with int/float support

mirguest:
BEGINRELEASENOTES
- A simple User Defined Data supporting int and float 
- A simple helper class is used to encode/decode the data in the collection. 
  - The fields and their offsets are declared explictly.
- Support two different styles
  - Each UserExt will only contain one element
  - UserExt could contain multiple elements. 
    - [ ] The getter of the number of elements is not implemented yet

ENDRELEASENOTES

This example is based on Frank's PR #114

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -415,3 +415,10 @@ datatypes :
       Members:
         - float value      // the single float value for extra user data
 
+  edm4hep::UserExt:
+      Description: "A simple struct with one user defined data"
+      Author : "Tao Lin"
+      VectorMembers:
+        - int valI // data int
+        - float valF // data float
+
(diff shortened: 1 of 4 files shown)


tmadlener:
Hi @mirguest, can you rebase this onto the new master, so that the github action workflows files are picked up?

mirguest:
@tmadlener , can you check whether it works or not? Thanks Thomas.
Tao

tmadlener:
Rebasing worked, thanks :)

However, there are now a few of the CI runs that are failing. From what it looks like we are hitting a (rather long-standing) bug in gcc here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282 (But I just see that you have fixed them already)

mirguest:
Hi @tmadlener ,
Yes, I have just fix the problem. I don't encounter such error when I use the macOS to compile EDM4hep. 
Tao

hegner:
I think you are right. So if that is enough for you, you can skip this WIP MR.

PR #118, opened 2021-07-22
Add github actions CI workflows 

tmadlener:
BEGINRELEASENOTES
- Add github actions CI workflows
- Fix issue with CI builds that build podio on the fly

ENDRELEASENOTES

Should we keep the travis CI workflows for now and remove them later, or should we migrate to github actions fully in this PR?

The linking problems that the "on-the-fly podio build" has, can easily be fixed by removing the unnecessary includes in the tricktrack test (`#include "podio/ROOTReader.h"` would need the `podio::podioRootIO` target to link against).

### DIFF
--- a/.github/workflows/key4hep.yml
+++ b/.github/workflows/key4hep.yml
@@ -1,0 +1,38 @@
+name: key4hep
+
+on: [push, pull_request]
+
+jobs:
+  build-and-test-key4hep:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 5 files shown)


vvolkl (approved):
Looks good, thanks. I don't think there is any reason to keep the travis configuration around, I'll remove it in another PR.

tmadlener:
I just realized the status badges in the README have the wrong url to the actual badge. Would need to be fixed as well in another PR.

PR #119, opened 2021-07-22
CI fixes: remove travis, update workflow names, fix badges

vvolkl:
BEGINRELEASENOTES
-  [ci] remove travis config (superseded by github actions)
-  [ci] rename workflows (`s/linux/lcg_linux/` as linux on its own is not very informative ) and fix badges.

ENDRELEASENOTES


vvolkl:
The doxygen generation still needs to be ported to github actions -> next PR.

README.md
```diff
@@ -1,7 +1,6 @@
 
-[![Build Status](https://travis-ci.com/key4hep/EDM4HEP.svg?branch=master)](https://travis-ci.com/key4hep/EDM4HEP)
-[![linux](https://github.com/key4hep/EDM4hep/workflows/linux.yml/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/linux.yml)
-[![key4hep](https://github.com/key4hep/EDM4hep/workflows/key4hep.yml/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/key4hep.yml)
+[![linux](https://github.com/key4hep/EDM4hep/workflows/lcg_linux/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/lcg_linux.yml)
+[![key4hep](https://github.com/key4hep/EDM4hep/workflows/key4hep_linux/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/key4hep_linux.yml)
```
tmadlener, line 3:
```suggestion
[![lcg_linux](https://github.com/key4hep/EDM4hep/actions/workflows/lcg_linux.yml/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/lcg_linux.yml)
[![key4hep_linux](https://github.com/key4hep/EDM4hep/actions/workflows/key4hep_linux.yml/badge.svg)](https://github.com/key4hep/EDM4hep/actions/workflows/key4hep_linux.yml)
```

I missed the `actions` in the original urls. Whether we want badges to just read `linux` or `lcg_linux` doesn't really matter. I simply copied the badge urls from the actions directly, where the filenames are used. For the other packages we only use `linux`.

tmadlener (changes requested):
I think the explicit naming of the workflows is a good idea. Do we also want the explicit names to appear in the README or are we happy with the non-explicit `linux` and `key4hep` there?

README.md
vvolkl: Would have worked in my version too, if you remove the `.yml` in the url, you don't need the `/actions/`

tmadlener: Ah, I didn't know that. TIL, thanks :)

tmadlener (approved):

Issue #112, opened 2021-03-30
[Fixed width types] Make all trivial members types fix width types


fdplacido:
As an example:
https://github.com/key4hep/EDM4hep/blob/9f83a691cb7bc8a6cfc42122866c615244133bc6/edm4hep.yaml#L174

It is declared as `unsigned long long`,  which is not guaranteed to always be 64 bits (it has to be at least 64 bits).
On conversion between EDM4hep and LCIO this may be an issue on bit manipulations like this case:

```cpp
unsigned long long combined_value =  edm_calohit.getCellID()
int* combined_value_int_ptr = reinterpret_cast<int*>(&combined_value);
lcio_calohit->setCellID0(combined_value_int_ptr[0]);
lcio_calohit->setCellID1(combined_value_int_ptr[1]);
```

Using `uint64_t` and `uint32_t` would make this more clear and not subject to possible errors with the size.

If EDM4hep is to be used in different architectures this would also be very desirable.

gaede:
Yes, it is a very good idea to use fixed width integers (and floats) in podio.

tmadlener:
As discussed during the [edm4hep meeting today](https://indico.cern.ch/event/1044074/) all integer and ideally also all floating point members should have fixed width types. For integers this should be a purely mechanical exercise of replacing them. For the floating point members some more work is necessary, and podio has to support them first.

- [ ] Switch integer types to fixed width types
- [ ] Switch floating point types to fixed width types
  - [ ] podio support for fixed width floating point types

fdplacido:
I would say we can close this issue since we have the integer fixed width types.
Having the floating point ones seems like a wider conversation or separate issue to have since its not as standard or defined as the integer ones.

tmadlener:
Yeah, I agree.

PR #125, opened 2021-10-12
Switch to fixed width integers everywhere

tmadlener:
BEGINRELEASENOTES
- Use fixed width integer types throughout the whole datamodel.

ENDRELEASENOTES

See #112 
Needs AIDAsoft/podio#234


vvolkl (approved):
Looks good to me. Maybe you can remove  the CI workflows with the old stacks that don't build their own updated podio?

tmadlener:
I kept the workflows for now, but changed the required tests to be the ones that build podio on the fly. Is there a way to keep the workflow files unchanged, but disable the actions from running?

PR #126, opened 2021-10-14
Fix test dependencies to allow for parallel running

tmadlener:
BEGINRELEASENOTES
- Fix test dependencies to make it possible to run them in parallel via the `-j` flag.

ENDRELEASENOTES


.github/workflows/key4hep_linux.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           cd build
           cmake .. -DCMAKE_CXX_STANDARD=17 \
             -DCMAKE_INSTALL_PREFIX=../install \
+            -DUSE_EXTERNAL_CATCH2=OFF \
```
vvolkl, line 25: I'd keep that on, I'll put a direct catch2 dependency in `key4hep-stack` to make sure it's there

vvolkl:
> https://github.com/key4hep/EDM4hep/pull/126/checks?check_run_id=3895745500#step:4:396
@tmadlener this could be an error with the setup script of the nightlies, I'll check and republish if necessary.

PR #130, opened 2021-11-16
[doc] update diagram for v0.4

vvolkl:
BEGINRELEASENOTES
- [doc] update diagram for v0.4

ENDRELEASENOTES

Adds the recently added TrackerHitPlane and Associations to the diagram. I'm omitting some association names to not make it too crowded, and the names follow a pattern anyway. 

Unrelated to this PR, while looking at the new diagram I started to wonder if podio should not just allow templated assocations between all types. I'll create an issue.

tmp

### DIFF
--- a/doc/edm4hep_diagram.svg
+++ b/doc/edm4hep_diagram.svg
@@ -13,7 +13,7 @@
    height="145.79602mm"
    width="286.02133mm"
    sodipodi:docname="edm4hep_diagram.svg"
-   inkscape:version="0.92.5 (2060ec1f9f, 2020-04-08)">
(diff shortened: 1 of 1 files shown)


tmadlener (commented):
Looks good to me.

I agree that a templated solution to all of these relations would be nice to have. On the other hand we might need to reconsider if we really need all these relations. Are these all present in LCIO as well? I suppose so, and we need them here for the converter to work, but it looks like almost everything now has an association to the MCParticle. Do we want that?

fdplacido:
The additions for the associations there were needed to run the full CLIC reconstruction, but in LCIO the mechanism for associations is different, so there could be more.
I will merge this PR since the changes are merged, so the repo and documentation can be updated accordingly.

PR #131, opened 2021-11-26
Switch to HepMC3 in tests

### LINKED ISSUE #60: HepMC3 support
Currently it seems that only `HepMC` up to version `2.06.10` is supported for the integration tests. Trying to build using `HepMC3` leads to several issues that are related to [differences between `HepMC` and `HepMC3`](http://hepmc.web.cern.ch/hepmc/differences.html).

Should we update to `HepMC3` (and state so in the documentation) as `HepMC` will no longer receive updates, or should we try and keep both versions for the integration tests?

The easiest way out is to update the documentation to make it explicit that the tests depend on `HepMC` and not `HepMC3`.

vvolkl:
BEGINRELEASENOTES
- switch to hepmc3 in tests

ENDRELEASENOTES

Fixes #60

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -74,13 +74,13 @@ endif()
 
 
 
-find_package(HepMC)
+find_package(HepMC3)
 find_package(HepPDT)
 
-if(HepMC_FOUND AND HepPDT_FOUND )
+if(HepMC3_FOUND AND HepPDT_FOUND )
   add_executable(edm4hep_testhepmc hepmc/edm4hep_testhepmc.cc)
-  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
-  target_link_libraries(edm4hep_testhepmc edm4hep podio::podioRootIO ${HEPPDT_LIBRARIES} ${HEPMC_LIBRARIES})
+  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -74,13 +74,13 @@ endif()
...
+find_package(HepMC3)
 find_package(HepPDT)
 
-if(HepMC_FOUND AND HepPDT_FOUND )
+if(HepMC3_FOUND AND HepPDT_FOUND )
   add_executable(edm4hep_testhepmc hepmc/edm4hep_testhepmc.cc)
-  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
-  target_link_libraries(edm4hep_testhepmc edm4hep podio::podioRootIO ${HEPPDT_LIBRARIES} ${HEPMC_LIBRARIES})
+  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
+  target_link_libraries(edm4hep_testhepmc edm4hep podio::podioRootIO ${HEPPDT_LIBRARIES} ${HEPMC3_LIBRARIES})
```
tmadlener, line 83: Minute detail here: Does HepMC3 export targets that could be used here?

vvolkl: Sadly no, see the upstream example: https://gitlab.cern.ch/hepmc/HepMC3/-/blob/master/examples/BasicExamples/CMakeLists.txt

tmadlener: Ah, pity.

tmadlener (approved):
Thanks for going through this exercise.

PR #133, opened 2021-12-09
Add association between ReconstructedParticle and Vertex

fdplacido:
BEGINRELEASENOTES
- Add association between ReconstructedParticle and Vertex

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -479,5 +479,5 @@ datatypes :
     Members:
       - float weight                      // weight of this association
     OneToOneRelations:
-     - edm4hep::ReconstructedParticle rec // reference to the track
-     - edm4hep::Vertex sim                // reference to the Monte-Carlo particle
+     - edm4hep::ReconstructedParticle rec // reference to the reconstructed particle
+     - edm4hep::Vertex sim                // reference to the vertex
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -480,4 +480,4 @@ datatypes :
       - float weight                      // weight of this association
     OneToOneRelations:
      - edm4hep::ReconstructedParticle rec // reference to the reconstructed particle
-     - edm4hep::Vertex sim                // reference to the vertex
+     - edm4hep::Vertex vertex             // reference to the vertex
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -472,3 +472,12 @@ datatypes :
     OneToOneRelations:
      - edm4hep::Track   rec         // reference to the track
      - edm4hep::MCParticle sim      // reference to the Monte-Carlo particle
+
+  edm4hep::RecoParticleVertexAssociation:
+    Description: "Association between a Reconstructed Particle and a Vertex"
+    Author : "Placido Fernandez Declara"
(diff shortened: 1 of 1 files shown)


tmadlener:
What is the use case for this? In principle a `Vertex` already has a relation to a `ReconstructedParticle` and vice versa.

https://github.com/key4hep/EDM4hep/blob/b1b0b2bb68009c253079f652423ede4b4dec6b5e/edm4hep.yaml#L382
and
https://github.com/key4hep/EDM4hep/blob/b1b0b2bb68009c253079f652423ede4b4dec6b5e/edm4hep.yaml#L399

andresailer:
a collection like this is created by LCFIPlus

https://github.com/lcfiplus/LCFIPlus/blob/278d7bb54e2602c2c0cd170c00f72fe302794bcf/src/LCIOStorer.cc#L954

tmadlener:
Ah yes, I see. I have opened an issue in podio to support arbitrary associations. Not entirely sure about the timeline though ;)

In principle I don't see anything that speaks against merging this now and then reworking things once (if) arbitrary associations arrive in podio.

edm4hep.yaml
```diff
@@ -471,4 +471,13 @@ datatypes :
...
+     - edm4hep::MCParticle sim      // reference to the Monte-Carlo particle
+
+  edm4hep::RecoParticleVertexAssociation:
+    Description: "Association between a Reconstructed Particle and a Vertex"
+    Author : "Placido Fernandez Declara"
+    Members:
+      - float weight                      // weight of this association
+    OneToOneRelations:
+     - edm4hep::ReconstructedParticle rec // reference to the track
+     - edm4hep::Vertex sim                // reference to the Monte-Carlo particle
```
andresailer, line 483:
```suggestion
     - edm4hep::Vertex sim                // reference to the vertex
```
Is `sim` needed as the name? I guess it is to keep the syntax the same in the k4LCIOReader converter? 
Because `sim` does not make sense in this case

edm4hep.yaml
```diff
@@ -471,4 +471,13 @@ datatypes :
...
\ No newline at end of file
+     - edm4hep::MCParticle sim      // reference to the Monte-Carlo particle
+
+  edm4hep::RecoParticleVertexAssociation:
+    Description: "Association between a Reconstructed Particle and a Vertex"
+    Author : "Placido Fernandez Declara"
+    Members:
+      - float weight                      // weight of this association
+    OneToOneRelations:
+     - edm4hep::ReconstructedParticle rec // reference to the track
```
andresailer, line 482:
```suggestion
     - edm4hep::ReconstructedParticle rec // reference to the reconstructed particle
```

fdplacido:
> Ah yes, I see. I have opened an issue in podio to support arbitrary associations. Not entirely sure about the timeline though ;)
> 
> In principle I don't see anything that speaks against merging this now and then reworking things once (if) arbitrary associations arrive in podio.

I would say we go ahead and merge this, since there are other association collections around already. All those would change once the arbitrary collections are in place.

vvolkl (approved):

PR #134, opened 2021-12-15
RecoParticleVertexAssociation: rename relation members for consistency between association

fdplacido:
BEGINRELEASENOTES
- Rename last association to be consistent with all other associations.

ENDRELEASENOTES

This allows to use any association in a generic (templates) way, so all the methods will be called the same.


vvolkl (approved):
Makes sense to me.

tmadlener:
Makes sense to me as well. Should we consider having more generic names than `rec` and `sim` then? I think all but this relation describe an association to an MC particle, so there it makes sense, but with the vertex it could be a bit confusing.

Depending on how many places this occurs in, one could also consider to have some `if constexpr` in place to handle the `RecoParticleVertexAssociation` just slightly different than the rest.

In any case this is likely a temporary solution until there is a more generic notion of relations/associations in podio, right?

fdplacido:
Maybe we should rename all association elements to "from" and "to", which seems the most generic thing to put. Since those are part of a collection named "*Association" should be pretty clear what that refers to, no?

tmadlener:
This would not break I/O, but probably some code that still has `rec` and `sim` as relation names. I have no strong preference on whether this should go in before a new tag. We will have to fix the naming of the code that uses this in any case.

tmadlener:
Can you quickly rebase, to get rid of the "out-of-date with the base branch"?

edm4hep.yaml
```diff
@@ -416,68 +416,68 @@ datatypes :
...
   edm4hep::MCRecoCaloAssociation:
     Description: "Association between a CaloHit and the corresponding simulated CaloHit"
     Author: "C. Bernet, B. Hegner"
     Members:
       - float weight                    // weight of this association
     OneToOneRelations:
-     - edm4hep::CalorimeterHit rec     // reference to the reconstructed hit
-     - edm4hep::SimCalorimeterHit sim  // reference to the simulated hit
+     - edm4hep::CalorimeterHit from     // reference to the reconstructed hit
+     - edm4hep::SimCalorimeterHit to    // reference to the simulated hit
```
tmadlener, line 429:
```suggestion
     - edm4hep::CalorimeterHit to     // reference to the reconstructed hit
     - edm4hep::SimCalorimeterHit from    // reference to the simulated hit
```
The name of the datatype suggests that the MC part is first, so I think it is more intuitive for it to be the "from" part of the association, right? A similar argument can be made for some of the other associations as well.

hegner:
Outdated and decided not to pick up

PR #137, opened 2022-02-22
Adding dNdX to edm4hep::Track

clementhelsens:
BEGINRELEASENOTES
- Adding dNdX to edm4hep::Track as discussed https://indico.cern.ch/event/1126908/
ENDRELEASENOTES


tmadlener:
Thanks. As also discussed during the meeting, leaving this open for a day or two to allow for some discussion.

My generalized proposal would have looked something like this:
1. Introduce a new component
```yaml
components:
  edm4hep::DxQuantity:
    Members:
      - int16_t type // flag identifying how to interpret the data
      - float value // the value
      - float error // the error
 ```
 2. Make the `Track` use that and remove `dEdx`
 ```yaml
datatypes:
  edm4hep::Track:
    Description: "Reconstructed track"
    Author : "F.Gaede, DESY"
    Members:
      - int32_t type                         //flagword that defines the type of track.Bits 16-31 are used internally
      - float chi2                       //Chi^2 of the track fit
      - int32_t ndf                          //number of degrees of freedom of the track fit
      - float radiusOfInnermostHit       //radius of the innermost hit that has been used in the track fit
    VectorMembers:
      - int32_t subDetectorHitNumbers        //number of hits in particular subdetectors.Check/set collection variable TrackSubdetectorNames for decoding the indices 
      - edm4hep::TrackState trackStates  //track states
      - edm4hep::DxQuantity dxQuantities // different measurements of dx quantities
    OneToManyRelations:
      - edm4hep::TrackerHit trackerHits  //hits that have been used to create this track
      - edm4hep::Track tracks            //tracks (segments) that have been combined to create this track
```
This would allow to store up to potentially 256 different dx categories and essentially a memory bound amount of actual dx measurements per track. The interface would become much less intuitive though and it only makes sense if there are actually more categories of dx measurements other than `dN` and `dE`.

clementhelsens:
Hi @tmadlener, I like your idea, as it allows for example storing various methods to compute dE/dX. Thus I suggest we proceed with your implementation and start using it in k4simDelphes

gaede:
Hi Thomas, nice idea.  Do we actually need an extra 16/8 bit type integer here or can we not make use of some bits in the type flag of the track ?  I would not expect more than 8 (3 bits) different types here.  Would safe a bit of disk space and also have better alignment...

tmadlener:
That would work. I can't really judge if having multiple different dx measurements makes sense for one track. If we only have one per track, then defining a dedicated component might also be a bit of an overkill, and we could get away with just proper naming.

clementhelsens:
@gaede , @tmadlener, I think we could have multiple definitions of dE/dX for a given track, meaning we can have different ways of computing the truncated mean. This is in my opinion a good argument to keep a vector member with arbitrary number of dq/dx

liwd2003:
1)	The number of dedx measurements used to calculate dedx of a track is less than or equal to the number of hits on the track because some dedx measurements will be dropped when the truncate mean method is applied. Both HitNumbers and number of dedx measurements should be member variables?

clementhelsens:
I believe we could even generalise @tmadlener suggestion with a generic component ```Quantity``` that would have an identifier, a value and an error, no need to make it specific to ```dx```

```cpp
components:
  edm4hep::Quantity:
    Members:
      - int16_t type // flag identifying how to interpret the data
      - float value // the value
      - float error // the error
```

andresailer (approved):
Approved in the meeting

tmadlener:
Merging this, since root's automatic schema evolution works for all relevant use cases for now. (As discussed during the [meeting on May 24](https://indico.cern.ch/event/1164601/)

tmadlener:
It is currently not clear to me why the CI builds against key4hep are failing for this. It looks like a dictionary mixup at first glance, but I am not sure that is actually the case, as there are also some unresolved symbol issues that seem to be unrelated to podio and EDM4hep. Additionally, only the RDataFrame test fails, but not the others which in principle also use the dictionaries and which are run in the same environment.

tmadlener:
I can also not seem to reproduce these problems locally, i.e. when I actually run what is run in CI in the same container as in the CI. @vvolkl any ideas what could be going on here?

tmadlener:
Thanks for fixing this.

PR #138, opened 2022-02-22
Adding Time in edm4hep::TrackState

clementhelsens:
As discussed several times, the time of the track is needed. For example, ACTS tracking code does. In addition, this time in ```edm4hep::TrackState``` can be used to obtain the time of the tracks at different points
BEGINRELEASENOTES
- TrackState: add time related entries to the covarianceMatrix

ENDRELEASENOTES

TODO:
- [x] @vvolkl to check if array size increase is possible with root's schema evolution
- [ ] cf. https://github.com/key4hep/EDM4hep/issues/154

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -74,8 +74,9 @@ components:
         - float omega // is the signed curvature of the track in [1/mm].
         - float Z0 // longitudinal impact parameter
         - float tanLambda // lambda is the dip angle of the track in r-z
+        - float time // time of the track at this trackstate
         - edm4hep::Vector3f referencePoint // Reference point of the track parameters, e.g. the origin at the IP, or the position  of the first/last hits or the entry point into the calorimeter.
-        - std::array<float, 15> covMatrix // lower triangular covariance matrix of the track parameters.  the order of parameters is  d0, phi, omega, z0, tan(lambda). the array is a row-major flattening of the matrix.
+        - std::array<float, 21> covMatrix // lower triangular covariance matrix of the track parameters.  the order of parameters is  d0, phi, omega, z0, tan(lambda), time. the array is a row-major flattening of the matrix.
       ExtraCode:
         declaration: "
         static const int AtOther = 0 ; // any location other than the ones defined below\n   


edm4hep.yaml
```diff
@@ -74,8 +74,9 @@ components:
         - float omega // is the signed curvature of the track in [1/mm].
         - float Z0 // longitudinal impact parameter
         - float tanLambda // lambda is the dip angle of the track in r-z
+	- float time // time of the track
         - edm4hep::Vector3f referencePoint // Reference point of the track parameters, e.g. the origin at the IP, or the position  of the first/last hits or the entry point into the calorimeter.
-        - std::array<float, 15> covMatrix // lower triangular covariance matrix of the track parameters.  the order of parameters is  d0, phi, omega, z0, tan(lambda). the array is a row-major flattening of the matrix.
+        - std::array<float, 21> covMatrix // lower triangular covariance matrix of the track parameters.  the order of parameters is  d0, phi, omega, z0, tan(lambda), time. the array is a row-major flattening of the matrix.
```
vvolkl, line 79: This will shuffle around the indices of covMatrix and break code that is currently using it ( not that much, but still):
https://github.com/cepc/CEPCSW/search?q=covMatrix&type=code
https://github.com/search?q=covMatrix+org%3Akey4hep&type=code

What about splitting this into two arrays? The full covariance matrix has to be constructed anyway for practical applications, so it makes no difference if that is done from covMat or covMat + covMatTimed

```suggestion
        - std::array<float, 15> covMatrix // lower triangular covariance matrix of the track parameters.  the order of parameters is  d0, phi, omega, z0, tan(lambda). the array is a row-major flattening of the matrix.
        - std::array<float, 7> covMatrixTimed // covariance matrix entries of the time-related track parameters.  the order of parameters is  sigma_t^2, d0, phi, omega, z0, tan(lambda)
```

andresailer: Actually as it is the lower triangle, it just adds indices at the end of the matrix, so not needed to split this in two, right?

edm4hep.yaml
```diff
@@ -74,8 +74,9 @@ components:
         - float omega // is the signed curvature of the track in [1/mm].
         - float Z0 // longitudinal impact parameter
         - float tanLambda // lambda is the dip angle of the track in r-z
+	- float time // time of the track
```
andresailer, line 77: Don't use tabs, replace by spaces.
```suggestion
	- float time // time of the track at this trackstate
```

edm4hep.yaml
clementhelsens: I agree that we do not need this complication of splitting the matrix as the existing indices will be unchanged.

vvolkl: That's true of course, I overlooked the **lower** triangle part!

fdplacido:
Just to keep track for k4MarlinWrapper when this gets merged, this will break here: https://github.com/key4hep/k4MarlinWrapper/blob/master/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L80

clementhelsens:
> Just to keep track for k4MarlinWrapper when this gets merged, this will break here: https://github.com/key4hep/k4MarlinWrapper/blob/master/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L80

that is typically a case to have ```auto``` no?

fdplacido:
> > Just to keep track for k4MarlinWrapper when this gets merged, this will break here: https://github.com/key4hep/k4MarlinWrapper/blob/master/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L80
> 
> that is typically a case to have `auto` no?

yes indeed. I would have to take care of `time` also, but `auto` will solve the issue for the matrix.

andresailer:
* Or ROOT only work around (ClassDef Version increase)?

tmadlener:
> * Or ROOT only work around (ClassDef Version increase)?

Our generated classes live entirely outside the ROOT `TObject` "ecosystem" and have no `ClassDef`.

PR #140, opened 2022-02-22
Remove the contributions member from the Clusters

### LINKED ISSUE #135: Hits and contributions should be aligned in some collections
To be aligned with LCIO, in some collections hits and contributions should be aligned: there must be same number of hits and contributions. But the EDM4hep interface encourages this to not be aligned.

Example:

In a `Cluster` one can `addToHits(hit)`:
```cpp
void Cluster::addToHits(::edm4hep::ConstCalorimeterHit component) {
  m_obj->m_hits->push_back(component);
  m_obj->data.hits_end++;
}
```
and also can `addToHitsContributions(contribution)`:
```cpp
void Cluster::addToHitContributions(float component) {
  m_obj->m_hitContributions->push_back(component);
  m_obj->data.hitContributions_end++;
}
```

But in LCIO, when adding a hit a contribution is always expected: https://github.com/iLCSoft/LCIO/blob/master/src/cpp/src/IMPL/ClusterImpl.cc#L133

---

- Both hit and hitContribution should go in sync
- A sensible default when a contribution is not indicated (i.e. 1.0)
- A method to add both at the same time should also be provided, this one would be fast and easy to implement in the templates.

tmadlener:
BEGINRELEASENOTES
- Remove the `hitContributions` vector member from the `edm4hep::Cluster` as it was discovered to be unused in LCIO.

ENDRELEASENOTES

- Fixes #135 
See notes of [Jan 25 meeting](https://indico.cern.ch/event/1117527/)


tmadlener:
Merging this, since root's automatic schema evolution works for all relevant use cases for now. (As discussed during the [meeting on May 24](https://indico.cern.ch/event/1164601/)

PR #143, opened 2022-03-04
Move the install destination of kinematics utility header

tmadlener:
BEGINRELEASENOTES
- Move the install destination of the `kinematics.h` utility header to be in `<prefix>/include/edm4hep/utils` (previously was in `<prefix>/include/utils/edm4hep`. This makes it better aligned with how namespacing is done, which is `edm4hep::utils`.
- Add minimal documentation for the `utils` folder.

ENDRELEASENOTES


tmadlener:
It looks like there are no users of this yet: https://github.com/search?q=edm4hep%2Fkinematics.h&type=code

I think we should be able to merge this now and include it in the new, upcoming tag.

vvolkl (approved):

Issue #144, opened 2022-03-15
EIC: utility functions for Vector{2,3}{f,d} etc


wdconinc:
In EIC we have developed several vector utility functions (some using c++20 concepts, but also as legacy) that we would be happy to upstream. In particular, see https://eicweb.phy.anl.gov/EIC/eicd/-/tree/master/utils/include/eicd for vector_utils.h and vector_utils_legacy.h. I am opening this as an issue to start a discussion. We can file a PR following up on this issue after some more discussion.

Primary use cases: calculation of theta, phi, eta from momentum and position vectors stored in the data model.

Alternatives considered: including this in the data model as ExtraCode adds overhead to the yaml description which is not pertinent to the data storage.

Note: because we cannot easily extend the edm4hep data model, we are reimplementing the same edm4hep data model in the eicd namespace. This is not desirable, but at this point in our transition to edm4hep unavoidable.

tmadlener:
Hi @wdoncic, We have started to put some utilities into [utils/include/edm4hep/utils](https://github.com/key4hep/EDM4hep/tree/master/utils/include/edm4hep/utils). Those will in the end be installed into `<prefix>/include/edm4hep/utils` and live in the `edm4hep::utils` namespace. Granted there is not a lot going on in there yet, but we would be very happy to accept additional functionality into that.

I would also be happy to have them with c++20 concepts already, but we will probably need a c++17 compatible version as well for the time being, or at least a version of the concepts that work with gcc8 as well.

wdconinc:
The restricting factor on our side has been the use within ROOT macros, which requires a legacy version anyway (at least until ROOT supports c++20, and even then it will take a while for that to be a reasonable requirement for a user base).

PR #149, opened 2022-03-23
Utilities for vector components

wdconinc:
As discussed in https://github.com/key4hep/EDM4hep/issues/144, these are some utility classes on vectors to ensure common operations are available. The utils with concepts are more maintainable, but a legacy version is attached too.

Credit to @sly2j

BEGINRELEASENOTES
- Add utilities to work with `Vector` component classes including
  - (`constexpr`) `operator` overloads for basic arithmetic with vectors
  - utility functionality that is commonly used with vectors
  - concepts to work with in generic vector related code.
- Include `edm4hep/utils/vector_utils.h` and work with functionality in `edm4hep::utils` namespace
  - This include will automatically use the concept powered version if supported by the compiler and fall back to a legacy implementation if not.

ENDRELEASENOTES

### DIFF
--- a/utils/include/edm4hep/utils/vector_utils.h
+++ b/utils/include/edm4hep/utils/vector_utils.h
@@ -208,7 +208,6 @@ namespace utils {
   }
 
 } // namespace utils
-} // namespace edm4hep
 
 template <edm4hep::Vector2D V>
 inline constexpr V operator+(const V& v1, const V& v2) {
@@ -267,6 +266,8 @@ template <edm4hep::VectorND V>
 inline constexpr V operator/(const V& v, const double d) {
   return (1. / d) * v;
 }
+
+} // namespace edm4hep
 
 #endif
 #endif
--- a/utils/include/edm4hep/utils/vector_utils.h
+++ b/utils/include/edm4hep/utils/vector_utils.h
@@ -12,24 +12,42 @@
 
 namespace edm4hep {
 
-template <class V> concept VectorHasX = requires(V v) { v.x; };
-template <class V> concept VectorHasY = requires(V v) { v.y; };
-template <class V> concept VectorHasZ = requires(V v) { v.z; };
-template <class V> concept VectorHasA = requires(V v) { v.a; };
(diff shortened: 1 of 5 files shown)


tmadlener (commented):
Hi @wdconinc, thanks for this. I like the concepts version of this :) I have a few minor comments inline. Some of them apply for the legacy version as well, although I didn't add them there a second time.

How much code would break on the EIC side if most of these utilities go into the `edm4hep::utils` namespace?

Also I assume the missing `Vector3d` overloads in the legacy version just mean that you have not yet come over that use case?

tmadlener:
Implemented the following changes wrt to the original PR:
- All `operators` are now marked `constexpr` to make them available at compile time (for whenever that might be handy)
- Brought the utility functionality into `namespace edm4hep::utils`.
- Expanded the legacy versions to work with all available vector types
- Added some basic tests for the `operator`s
- Marked all constructors and extra code for the vector components as `constexpr` to make them work with the rest

Also brought this back to the latest status of the current main branch (mainly clang-format).

sly2j:
From my side, I like the updated version. I don't see a problem with updating the namespaces in the EIC software.

wdconinc:
Looks good. I just checked with our local version and we don't have additional utils that aren't already in here.

utils/include/edm4hep/utils/vector_utils.h
```diff
@@ -0,0 +1,270 @@
...
+  using ValueType = typename detail::ValueTypeHelper<V>::type;
+
+  // inline edm4hep::Vector2f VectorFromPolar(const double r, const double theta)
+  // {
+  //  return {r * sin(theta), r * cos(theta)};
+  //}
+
+  template <Vector3D V = edm4hep::Vector3f>
+  V sphericalToVector(const double r, const double theta, const double phi) {
+    using FloatType = decltype(vector_x(V()));
```
tmadlener, line 123:
```suggestion
    using FloatType = ValueType<V>;
```
Since this is now a thing.

Issue #145, opened 2022-03-15
MCParticle::getEnergy may use std::hypot (c++11) for better numerical control


wdconinc:
Currently https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L162 defines:
```cpp
      double getEnergy() const { return sqrt( getMomentum()[0]*getMomentum()[0]+getMomentum()[1]*getMomentum()[1]+\n
                                        getMomentum()[2]*getMomentum()[2] + getMass()*getMass()  )  ;}
```
but numerical stability is likely better with
```cpp
      double getEnergy() const { return std::hypot(getMomentum()[0], getMomentum()[1], getMomentum()[2], getMass()); }
```

Other considerations: compatiblity with pre-c++11 would preclude this, and a reimplementation in a utility function could have backwards compatibility implications.

tmadlener:
Since podio already requires at least c++17, I don't think trying to be compatible with pre-c++11 is a concern.

hegner:
I agree. please go on and make a PR. Thanks!

andresailer:
As pointed out in #150, there is no 4 argument hypot, so this is a no-can-do. Closing this issue

PR #150, opened 2022-03-23
Prefer std::hypot over sqrt of sum of squares

wdconinc:
Though overflows are unlikely to occur here, this is likely more numerically stable for finite arguments that do occur in NHEP. "Implementations usually guarantee precision of less than 1 ulp" https://en.cppreference.com/w/cpp/numeric/math/hypot.

Ref. https://github.com/key4hep/EDM4hep/issues/145.

BEGINRELEASENOTES
- std::hypot for sqrt of sum of squares in geEnergy and rho

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -159,8 +159,7 @@ datatypes :
       static const int BITStopped = 24 ;    \n
       static const int BITOverlay = 23 ;    \n
       /// return energy computed from momentum and mass \n
-      double getEnergy() const { return sqrt( getMomentum()[0]*getMomentum()[0]+getMomentum()[1]*getMomentum()[1]+\n
-                                        getMomentum()[2]*getMomentum()[2] + getMass()*getMass()  )  ;} \n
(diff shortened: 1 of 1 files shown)


edm4hep.yaml
```diff
@@ -159,8 +159,7 @@ datatypes :
       static const int BITStopped = 24 ;    \n
       static const int BITOverlay = 23 ;    \n
       /// return energy computed from momentum and mass \n
-      double getEnergy() const { return sqrt( getMomentum()[0]*getMomentum()[0]+getMomentum()[1]*getMomentum()[1]+\n
-                                        getMomentum()[2]*getMomentum()[2] + getMass()*getMass()  )  ;} \n
+      double getEnergy() const { return std::hypot(getMomentum()[0], getMomentum()[1], getMomentum()[2], getMass()) ; } \n
```
tmadlener, line 162: There doesn't seem to be a 4 argument overload for `std::hypot`

wdconinc: Well shoot

wdconinc:
Without a four argument overload this is probably just better left unchanged. I don't know why this didn't fail earlier in our code. And now of course I cannnot make it work anymore either...

PR #158, opened 2022-06-15
Expand tests with more classes

stoimenova:
Since the write_events and read_events test files only contained examples from four classes, I expanded them with two more.

BEGINRELEASENOTES
- Expanded write_events and read_events test with two new classes
ENDRELEASENOTES

### DIFF
--- a/test/read_events.h
+++ b/test/read_events.h
@@ -214,7 +214,7 @@ void processEvent(podio::EventStore& store, bool verboser, unsigned eventNum) {
  //  }
 
 auto& evtMD = store.getEventMetaData() ;
-   std::string evtType = evtMD.getStringVal( "EventType" ) ;
+   const auto& evtType = evtMD.getValue<std::string>( "EventType" ) ;
     std::cout<<"Event Type: "<<evtType<<std::endl;
 
 
--- a/test/read_events.h
+++ b/test/read_events.h
@@ -6,6 +6,8 @@
 #include "edm4hep/SimTrackerHitCollection.h"
 #include "edm4hep/CaloHitContributionCollection.h"
 #include "edm4hep/SimCalorimeterHitCollection.h"
+#include "edm4hep/TPCHitCollection.h"
+#include "edm4hep/TrackerHitPlaneCollection.h"
 
 // podio specific includes
 #include "podio/EventStore.h"
@@ -21,6 +23,8 @@ void processEvent(podio::EventStore& store, bool verboser, unsigned eventNum) {
   auto& sths   = store.get<edm4hep::SimTrackerHitCollection>("SimTrackerHits");
   auto& schs   = store.get<edm4hep::SimCalorimeterHitCollection>("SimCalorimeterHits");
   auto& sccons = store.get<edm4hep::CaloHitContributionCollection>("SimCalorimeterHitContributions");
+  auto& tpchs  = store.get<edm4hep::TPCHitCollection>("TPCHits");
+  auto& thps   = store.get<edm4hep::TrackerHitPlaneCollection>("TrackerHitPlanes");
 
 
   if( mcps.isValid() ){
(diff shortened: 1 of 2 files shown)


test/write_events.h
```diff
@@ -181,6 +190,51 @@ void write(std::string outfilename) {
...
         << schs << std::endl ;
 
+
+    //===============================================================================
+    // write some TPCHits:
+    int ntpch = 5 ;
+    for(int j=0 ; j<ntpch ; ++j){
+      auto tpch1 = tpchs.create() ;
+      tpch1.setCellID( 0xabadcaffee ) ;
+      tpch1.setTime( j*3. ) ;
```
tmadlener, line 200:
```suggestion
      tpch1.setTime( j*3.f ) ;
```
I think if you set the time here as `float` (instead of as an implicit `double`) you can simplify the reading side of this

test/write_events.h
```diff
@@ -57,56 +66,56 @@ void write(std::string outfilename) {
 
     auto mcp1 = mcps.create();
     mcp1.setPDG( 2212 ) ;
-    mcp1.setMass( 0.938 ) ;
+    mcp1.setMass( 0.938d ) ;
```
vvolkl, line 69:
```suggestion
    mcp1.setMass( 0.938f ) ;
```
Floating point numbers are implicitly of type double. The `d` suffix is a gcc extension, thus the build failures with other compilers.

vvolkl (changes requested):
@stoimenova See comment above

test/read_events.h
```diff
@@ -153,12 +157,67 @@ void processEvent(podio::EventStore& store, bool verboser, unsigned eventNum) {
...
+ }
+
  // //===============================================================================
  //  if( sccons.isValid() ){
  //  } else {
  //    throw std::runtime_error("Collection 'SimCalorimeterHitContributions' should be present");
  //  }
 
+auto& evtMD = store.getEventMetaData() ;
+   std::string evtType = evtMD.getStringVal( "EventType" ) ;
```
tmadlener, line 217:
```suggestion
   const auto& evtType = evtMD.getValue<std::string>( "EventType" ) ;
```
Templated access has been added in AIDASoft/podio#262 . In the course of that this interface has been marked as deprecated. If you build against a recent enough podio version you should be able to pick up the newer interface.

vvolkl (approved):

test/write_events.h
```diff
@@ -181,9 +190,62 @@ void write(std::string outfilename) {
...
+
+    auto& evtMD = store.getEventMetaData() ;
+    evtMD.setValue( "EventType" , "test") ;   
 
     writer.writeEvent();
+
+
+
+
+
```
tmadlener, line 248:
```suggestion
```

test/write_events.h
```diff
@@ -181,9 +190,62 @@ void write(std::string outfilename) {
...
+      auto tpch2 = tpchs.create() ;
+      tpch2.setCellID( 0xcaffeebabe ) ;
+      tpch2.setTime( j*0.3f ) ;
+      tpch2.setCharge( -j*2.f ) ;
+    }
+
+    std::cout <<"\n collection: " << "Time Projection Chamber Hits" << " of type " << tpchs.getValueTypeName() << "\n\n"
+	      << tpchs << std::endl ;
+ 
+
```
tmadlener, line 212:
```suggestion
```

test/write_events.h
```diff
@@ -181,9 +190,62 @@ void write(std::string outfilename) {
...
+      auto thp2 = thps.create() ;
+      thp2.setCellID( 0xcaffeebabe ) ;
+      thp2.setTime( j*0.3f ) ;
+      thp2.setEDep( j * 0.0000031f ) ;
+      thp2.setPosition( { -j*2. , -j*3. , -j*5. } ) ;
+    }
+
+
+
```
tmadlener, line 232:
```suggestion
```

test/write_events.h
```diff
@@ -181,9 +190,62 @@ void write(std::string outfilename) {
...
+    std::cout <<"\n collection: " << "Tracker Hit Planes" << " of type " << thps.getValueTypeName() << "\n\n"
+	      << thps << std::endl ;
+ 
+
+
```
tmadlener, line 237:
```suggestion
```

tmadlener:
Key4hep CI currently failing in the downstream tests due to not recent enough podio version. Merging this regardless.

PR #160, opened 2022-06-23
update edm4hep.yaml doc

andiwand:
updating the doc for `TrackerHitPlane` after clarifying the meaning of `u` and `v`

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -329,8 +329,8 @@ datatypes :
       - float time                     //time of the hit.
       - float eDep                     //energy deposited on the hit [GeV].
       - float eDepError                //error measured on EDep [GeV].
-      - edm4hep::Vector2f u            //measurement direction vector, u lies in the x-y plane
-      - edm4hep::Vector2f v            //measurement direction vector, v is along z
+      - edm4hep::Vector2f u            //measurement direction vector for X in global spherical coordinates with R=1 -> (theta, phi)
+      - edm4hep::Vector2f v            //measurement direction vector for Y in global spherical coordinates with R=1 -> (theta, phi)
       - float du                       //measurement error along the direction
       - float dv                       //measurement error along the direction
       - edm4hep::Vector3d position     //hit position in [mm].


edm4hep.yaml
```diff
@@ -329,8 +329,8 @@ datatypes :
       - float time                     //time of the hit.
       - float eDep                     //energy deposited on the hit [GeV].
       - float eDepError                //error measured on EDep [GeV].
-      - edm4hep::Vector2f u            //measurement direction vector, u lies in the x-y plane
-      - edm4hep::Vector2f v            //measurement direction vector, v is along z
+      - edm4hep::Vector2f u            //measurement direction vector for X in global spherical coordinates with R=1 -> (theta, phi)
+      - edm4hep::Vector2f v            //measurement direction vector for Y in global spherical coordinates with R=1 -> (theta, phi)
       - float du                       //measurement error along the direction
       - float dv                       //measurement error along the direction
       - edm4hep::Vector3d position     //hit position in [mm].
```
andiwand, line 336: if I understand it right U defines the X axis and V defines the Y axis for `position`. the Z component will be ignored then?

what is also not clear to me: why is there a `du` and `dv` if there is also a `covMatrix` that describes the covariance of `position`

gaede: This class has been copied from LCIO - see here for a more complete documentation: [https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHitPlane.html](https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHitPlane.html).
The measurement directions `u,v` are the two directions of the measurement surface in global (polar) coordinates. The `du, dv` are the resolutions along these directions, i.e. the *point resolution*.  So `u,v` are not really X and Y.
The `covMatrix` is given in the global cartesian coordinate and can be given for all types of `TrackerHits`

andiwand: thank you @gaede ! will the `position` contain then the `U` `V` values in form of `X` `Y`? this is where I get confused because you said `covMatrix` will be in global cartesian coordinates

gaede: No, `u,v` really should be in `(theta, phi)`, i.e spherical (and not polar coordinates). The global cartesian covariance matrix is a left-over from a time when there was only the generic `TrackerHit` in LCIO.
One could think about dropping this from EDM4hep...

andiwand: but `U` and `V` will only span the measurement direction for the plane, right? as you said in `(theta, phi)` global spherical coordinates.

but where do you then store the actual local measurement values for `U` and `V`? will they go into the `position`'s `X` and `Y`? and what happens to the `Z` component?

gaede: Only global coordinates are used here. If local coordinates are needed they can be computed from the surface class used in the tracking.

paulgessinger: So IIUC, `u,v` is always global, `du,dv` are in effect the local uncertainties (diagonal cov entries, correlations are dropped). OTOH `xyz` and the covariance matrix are fully global. Would be good to clarify this.

> This class has been copied from LCIO - see here for a more complete documentation: https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHitPlane.html.

That is a good resource, but this information has to move into this repository at some point.

gaede: Agreed, the documentation needs to be cleaned up and extended.

Not sure I understand your comment about local uncertainties and their correlation being dropped. There should be none - at least that is what is assumed here.

paulgessinger: You can recover the off-diagonal covariance entries from the global position and global covariance (as well as `du,dv` I suppose), but they're not directly stored (while `du,dv` are). That was all.

paulgessinger: Ah, I actually thought it was the actual diagonal covariance matrix elements of the measurement.

andiwand: actually you cannot calculate the local coordinates because it would require the global cartesian coordinates of the origin of the detector module that was hit, right?

PR #162, opened 2022-07-09
[cmake] set EDM4HEP_SET_RPATH ON by default; deduplicate RPATH settings

vvolkl:
BEGINRELEASENOTES
- [[cmake] set EDM4HEP_SET_RPATH ON by default
- [cmake] deduplicate RPATH settings

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ endif()
 
 message (STATUS "C++ standard: ${CMAKE_CXX_STANDARD}")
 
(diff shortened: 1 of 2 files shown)


tmadlener (approved):
This looks good to me. Out of curiosity, did we ever use the old option somewhere in our builds explicitly?

vvolkl:
No, but it interferred with the RPATH settings that are now mandatory to have when building against the key4hep setup.sh, which no longer sets the LD_LIBRARY_PATH for most things.

PR #163, opened 2022-07-19
Adding dummy EDM4hep .root -> Phoenix .json converter

kjvbrt:
Adding dummy EDM4hep -> Phoenix json converter. So far, only clusters -> jets are converted.


tmadlener:
Hi @kjvbrt,

this looks very interesting and useful. I am wondering whether we should not leverage the template engine facilities of podio to automatically generate the necessary conversion code. Would something like simply putting all available information from the datamodel into a json dictionary be enough, or does phoenix expect specific information for each type? E.g. cylindrical coordinates instead of cartesian for positions, etc.?

vvolkl:
@tmadlener, interesting idea, but I guess podio would need to introduce the notion of a "displayable type/member"? As most of this logic concerns the datamodel content.

tmadlener:
I think simply dumping to json should not be too hard from a podio perspective. I can't say whether the necessary transformations are then easier in the Phoenix loader or in a dedicated converter. Maybe you have a better feeling about that. Would relation information need to be preserved somehow as well for this to work? (e.g. tracker hits in a track).

kjvbrt:
I'm not sure either, but having standard JSON output would be better from the interoperability perspective.

The Phoenix recognizes only relation between tracks and clusters (electrons, muons) at the moment.

tmadlener:
Let me try to put together a prototype JSON output for podio to see if that can be useful for this purpose and what would still be necessary for Phoenix.

tmadlener:
@kjvbrt I have a prototype version of JSON support in AIDASoft/podio#312 if you want to give it a try. I haven't fully tested it myself, but I think it should work in principle. In case of any problems we can also discuss on that PR.

kjvbrt:
Using new ability of podio to generate JSON results in a linker error.

tmadlener:
I think the problem is that you are only linking the executable against `nlohmann_json`, but this should actually go to the core edm4hep library, i.e. something like the following
```cmake
find_package(nlohmann_json 3.10)
if (nlohmann_json_FOUND)
  target_compile_definitions(edm4hep PUBLIC PODIO_JSON_OUTPUT)
  target_link_libraries(edm4hep PUBLIC nlohmann_json::nlohmann_json)
endif()
```

should be added here:
https://github.com/key4hep/EDM4hep/blob/fb3d92b3f6c47584f3d3256706d400bde5344ecc/edm4hep/CMakeLists.txt#L6

kjvbrt:
That is the trick. Now it compiles and runs.

tools/src/edm4hep2json.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+// ROOT
+#include "podio/ROOTReader.h"
+
+// *nix
+#include <getopt.h>
+
+void printHelp() {
+  std::cout << "Usage:\n"
+            << "  -i/--in-file            input file path\n"
```
tmadlener, line 12: Not sure if `getopt` supports this, but this seems to be the only required input parameter, can it be made to one that does not require a flag? i.e. usage should look something like

`edm4hep2json [flag-args] inputfile`

tools/src/edm4hep2json.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+  if (inFile.empty()) {
+    std::cout << "ERROR: Input .root file not provided!" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  if (requestedCollections.empty()) {
+    requestedCollections =
+        "GenParticles,BuildUpVertices,SiTracks,PandoraClusters,VertexJets";
+    std::cout << "WARNING: Using default collection to convert:\n"
```
tmadlener, line 80: If this list is what works in > 90% of the cases, I would remove this warning (or at the very least demote it to a debug message, that only shows up with `--verbose`).

tools/src/edm4hep2json.cxx
```diff
@@ -0,0 +1,97 @@
...
+              << "         " << requestedCollections << std::endl;
+  }
+
+  if (outFile.empty()) {
+    outFile = "events.edm4hep.json";
+    std::cout << "WARNING: Using default output file path:\n"
```
tmadlener, line 86: Also this is not really a warning if the default is valid and appropriately chosen. In principle this could also be derived from the input filename.

tools/include/edm4hep2json.hxx
```diff
@@ -0,0 +1,173 @@
...
+
+// STL
+#include <vector>
+#include <iostream>
+#include <fstream>
+#include <exception>
+#include <cassert>
+
+
+nlohmann::json processEvent(podio::EventStore& store,
```
tmadlener, line 26:
```suggestion
template<typename PodioStoreT>
nlohmann::json processEvent(const PodioStoreT& store,
```
This should automatically make it also work with the `Frame`.

tools/include/edm4hep2json.hxx
```diff
@@ -0,0 +1,173 @@
...
+      };
+      jsonDict.insert(jsonColl.begin(), jsonColl.end());
+    }
+  }
+
+  return jsonDict;
+}
+
+
+void printCollTypes(podio::EventStore& store,
```
tmadlener, line 91:
```suggestion
template<typename PodioStoreT>
void printCollTypes(const PodioStore& store,
```
same here

tmadlener (commented):
This looks good to me, I only have a few minor comments.

How much "legacy" EDM4hep input is this supposed to work with? If the plan is to mainly work with newer files, then I would suggest to go to `podio::Frame` usage (including its readers) directly.

tools/include/edm4hep2json.hxx
kjvbrt: Thanks for the suggestion

kjvbrt:
At the moment we aim this to be used for the tutorial. I would use json only as the intermediate file format, not to store any data in it.

PR #165, opened 2022-08-02
EDM4DIS derived datamodel

wdconinc:
This adds a very light-weight EDM4DIS data model which combines a new data type with EDM4hep data types.

BEGINRELEASENOTES
- test and example for data model derived from EDM4hep

ENDRELEASENOTES

Needs #164 and AIDASoft/podio#317

### DIFF
--- a/test/downstream-project-cmake-test/CMakeLists.txt
+++ b/test/downstream-project-cmake-test/CMakeLists.txt
@@ -28,9 +28,9 @@ else()
   PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}"
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
-target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
+  target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
 
-PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
+  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
 endif()
--- a/test/downstream-project-cmake-test/CMakeLists.txt
+++ b/test/downstream-project-cmake-test/CMakeLists.txt
@@ -29,7 +29,7 @@ else()
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
 
-  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" ${CMAKE_CURRENT_BINARY_DIR}/src/selection.xml
+  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
 endif()
--- a/test/downstream-project-cmake-test/CMakeLists.txt
+++ b/test/downstream-project-cmake-test/CMakeLists.txt
@@ -28,7 +28,7 @@ else()
   PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}"
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
-
+target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
   PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
     OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
   )
--- a/test/downstream-project-cmake-test/CMakeLists.txt
+++ b/test/downstream-project-cmake-test/CMakeLists.txt
@@ -13,5 +13,34 @@ else()
   message(STATUS "The current exported EDM4HEP version is " ${EDM4HEP_VERSION})
 endif()
 
+# Use the EDM4HEP_DATA_DIR to derive extended data model
+if (${EDM4HEP_DATA_DIR})
+  message(STATUS "Current EDM4HEP data model is available at ${EDM4HEP_DATA_DIR}.")
+  PODIO_GENERATE_DATAMODEL(EDM4DIS edm4dis.yaml headers sources
+    UPSTREAM_EDM edm4hep:${EDM4HEP_DATA_DIR}/edm4hep.yaml
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+  )
+  add_library(EDM4DIS SHARED
+    ${sources}
+  )
+
+  target_link_libraries(EDM4DIS
+    PUBLIC EDM4HEP::edm4hep
+    PUBLIC podio::podio
+  )
+  target_include_directories(EDM4DIS
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+    $<INSTALL_INTERFACE:include>
+  )
+
+  PODIO_GENERATE_DICTIONARY(EDM4DIS ${headers} 
+    SELECTION ${CMAKE_CURRENT_BINARY_DIR}/src/selection.xml
+    OPTIONS --library ${CMAKE_SHARED_LIBRARY_PREFIX}EDM4DIS${CMAKE_SHARED_LIBRARY_SUFFIX}
+  )
+  set_target_properties(edm4dis-dictgen PROPERTIES EXCLUDE_FROM_ALL TRUE)
+else()
+  message(WARNING "Unable to extend data model since no installed yaml file found.")
+endif()
+
 add_executable(appUsingEDM4hep main.cxx)
 target_link_libraries(appUsingEDM4hep EDM4HEP::edm4hep EDM4HEP::utils)
--- a/test/downstream-project-cmake-test/CMakeLists.txt
+++ b/test/downstream-project-cmake-test/CMakeLists.txt
@@ -15,7 +15,7 @@ endif()
 
 # Use the EDM4HEP_DATA_DIR to derive extended data model
 message(STATUS "${EDM4HEP_DATA_DIR}")
-if (${EDM4HEP_DATA_DIR} STREQUAL "")
+if ("${EDM4HEP_DATA_DIR}" STREQUAL "")
   message(WARNING "Unable to extend data model since no installed yaml file found.")
 else()
   message(STATUS "Current EDM4HEP data model is available at ${EDM4HEP_DATA_DIR}.")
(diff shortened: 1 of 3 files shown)


test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,35 @@ else()
...
+  target_include_directories(edm4dis
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+    $<INSTALL_INTERFACE:include>
+  )
+
+  PODIO_GENERATE_DICTIONARY(edm4dis ${headers} 
+    SELECTION ${CMAKE_CURRENT_BINARY_DIR}/src/selection.xml
+    OPTIONS --library ${CMAKE_SHARED_LIBRARY_PREFIX}edm4dis${CMAKE_SHARED_LIBRARY_SUFFIX}
+  )
+  set_target_properties(edm4dis-dictgen PROPERTIES EXCLUDE_FROM_ALL TRUE)
```
tmadlener, line 43: In this case it probably won't make too much of a difference, since you have to link against edm4hep in any case, but just in case you are not aware of this, there is a `PODIO_ADD_DATAMODEL_CORE_LIB` and a `PODIO_ADD_ROOT_IO_DICT` macro that handles this, see, e.g. how they are used in edm4hep:
https://github.com/key4hep/EDM4hep/blob/fb3d92b3f6c47584f3d3256706d400bde5344ecc/edm4hep/CMakeLists.txt#L4-L8

Definitions/documentation:
- [`PODIO_ADD_DATAMODEL_CORE_LIB`](https://github.com/AIDASoft/podio/blob/e576e36c8bd9dd90ec27bf96166a4d5b891148e6/cmake/podioMacros.cmake#L204)
- [`PODIO_ADD_ROOT_IO_DICT`](https://github.com/AIDASoft/podio/blob/e576e36c8bd9dd90ec27bf96166a4d5b891148e6/cmake/podioMacros.cmake#L244)

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,35 @@ else()
   message(STATUS "The current exported EDM4HEP version is " ${EDM4HEP_VERSION})
 endif()
 
+# Use the EDM4HEP_DATA_DIR to derive extended data model
+message(STATUS "${EDM4HEP_DATA_DIR}")
+if ("${EDM4HEP_DATA_DIR}" STREQUAL "")
+  message(WARNING "Unable to extend data model since no installed yaml file found.")
```
tmadlener, line 19:
```suggestion
  message(FATAL_ERROR "Unable to extend data model since no installed yaml file found.")
```

For the tests this should fail I think to make sure that we can actually build against this.

wdconinc: The main reason for not having it be FATAL_ERROR was that I wasn't sure of the timeline for the required changes to podio and EDM4hep. It seems that's moving forward quickly though so we can expect this to pass tests again soon.

test/downstream-project-cmake-test/CMakeLists.txt
wdconinc: Hmm, tried that, but this now fails to include the generated header directory in the build directory on the compilation line.

wdconinc:
```console
[ 12%] Building CXX object CMakeFiles/edm4dis.dir/src/InclusiveKinematics.cc.o
/home/wdconinc/git/EDM4hep/test/downstream-project-cmake-test/build/src/InclusiveKinematics.cc:4:10: fatal error: edm4dis/InclusiveKinematics.h: No such file or directory
    4 | #include "edm4dis/InclusiveKinematics.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

wdconinc: Marked as draft since it's not working right now.

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,22 @@ else()
...
+  message(FATAL_ERROR "Unable to extend data model since no installed yaml file found.")
+else()
+  message(STATUS "Current EDM4HEP data model is available at ${EDM4HEP_DATA_DIR}.")
+  PODIO_GENERATE_DATAMODEL(edm4dis edm4dis.yaml headers sources
+    UPSTREAM_EDM edm4hep:${EDM4HEP_DATA_DIR}/edm4hep.yaml
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+  )
+
+  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}")
```
tmadlener, line 28:
```suggestion
  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}" OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}) 
```
I think this should fix the issue here.

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,22 @@ else()
...
+  message(STATUS "Current EDM4HEP data model is available at ${EDM4HEP_DATA_DIR}.")
+  PODIO_GENERATE_DATAMODEL(edm4dis edm4dis.yaml headers sources
+    UPSTREAM_EDM edm4hep:${EDM4HEP_DATA_DIR}/edm4hep.yaml
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+  )
+
+  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}") 
+  
+  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml)
```
tmadlener, line 30:
```suggestion
  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}) 
```
I think also this one needs an explicit output directory if it is not the current source directory. Potentially also the ˋselection.xmlˋ path assumes that and would need an adaption in this case.

I just realize that these macros are not as ergonomic as they could be outside of the defaults.

wdconinc:
This now works as intended. Marked as ready for review.

- I wish the default link target would be `EDM4DIS::edm4dis` instead of `edm4dis` in the last line,
```cmake
target_link_libraries(appUsingEDM4hep EDM4HEP::edm4hep EDM4HEP::utils edm4dis)
```

wdconinc:
I just noticed the following in the cmake output:
```console
-- Not adding the ROOT dictionary to the targets because the selection.xml has not been generated
```

wdconinc:
It seem SELECTION_XML is always searched for under OUTPUT_FOLDER.

wdconinc:
Further analysis indicates that this is caused by not passing the correct `selection.xml` path to the `PODIO_GENERATE_DICTIONARY` macro. This is fixed by https://github.com/AIDASoft/podio/pull/318. With that small patch, this now builds the derived data model correctly, including dictionaries `libedm4disDict.so`.

tmadlener:
> - I wish the default link target would be EDM4DIS::edm4dis instead of edm4dis in the last line,
 > ```
 > target_link_libraries(appUsingEDM4hep EDM4HEP::edm4hep EDM4HEP::utils edm4dis)
 > ```
 
 For this to work you would have to introduce an alias, similar to what is done here:
 https://github.com/key4hep/EDM4hep/blob/ffcdb0b34766e444d145367966037852f7692431/edm4hep/CMakeLists.txt#L9
 
 This is not something that we can (or want) to do on the podio side, because there might be different conventions at play. Additionally, you also should make sure that the alias that you introduce here is the same as the namespaced targed in downstream usage.

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,26 @@ else()
...
+  PODIO_GENERATE_DATAMODEL(edm4dis edm4dis.yaml headers sources
+    UPSTREAM_EDM edm4hep:${EDM4HEP_DATA_DIR}/edm4hep.yaml
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+  )
+
+  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}"
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+  )
+
```
tmadlener, line 31:
```suggestion
target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
```
The podio macro that builds the core lib does not know about the upstream model so it has to be explicitly linked against that.

tmadlener:
Key4hep build (nightlies) is currently failing because it does not yet have a podio version with the required PR merged.

tmadlener:
Merging already now, even though the Key4hep CI builds are failing. They currently don't have a recent enough version of podio, but builds on master are covered by the LCG builds in any case.

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,26 @@ else()
...
+    UPSTREAM_EDM edm4hep:${EDM4HEP_DATA_DIR}/edm4hep.yaml
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+  )
+
+  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}"
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+  )
+target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
+  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
```
tmadlener, line 32:
```suggestion

PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
```

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -13,5 +13,27 @@ else()
...
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+  )
+
+  PODIO_ADD_DATAMODEL_CORE_LIB(edm4dis "${headers}" "${sources}"
+    OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
+  )
+target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)
+
+PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
```
tmadlener, line 33:
```suggestion
  target_link_libraries(edm4dis PUBLIC EDM4HEP::edm4hep)

  PODIO_ADD_ROOT_IO_DICT(edm4disDict edm4dis "${headers}" src/selection.xml
```

Issue #161, opened 2022-06-24
Add local Measurement{1,2,3}D classes


paulgessinger:
As discussed in the AIDAInnova hackathon during during CW25 2022, we should investigate providing classes for proper local measurements connected to a geometry description. The conclusion was to skip `TrackerHitPlane` in direct ACTS output, and provide a `MeasurementXD` to `TrackerHitPlane` conversion mechanism instead.

Can we use this issue to discuss this? @gaede @andresailer @vvolkl @andiwand @asalzburger @tmadlener

tmadlener:
As mentioned during the hackathon already the main concerns from EDM4hep side are the following:
1. Introducing a "data sink" into which users will just start to dump anything, instead of putting it into the EDM properly. We had quite some discussion on how to allow users to store some additional data that does not yet fit the EDM and came up with, what we call `UserDataCollection` in podio: [documentation](https://github.com/AIDASoft/podio/blob/master/doc/userdata.md),  [UserDataCollection.h](https://github.com/AIDASoft/podio/blob/master/include/podio/UserDataCollection.h)
2. Tying the `Measurement` classes too tightly to ACTS (or any other component that could make use of them).

For the first one I have no clever solution yet, other than discouraging usage via documentation. The second one should be a bit easier to handle by providing some access functionality via some user / ACTS provided enum (or comparable). This would essentially work by mapping enum values to indices, as the access via integer indices will always be possible. I would have to actually sit down and try to come up with an implementation to see where this might fail.

One thing that is not yet entirely clear to me is whether ACTS would also want to persist the `Measurement` to file or whether it would mainly be used for transient access?

A final thought about the actual implementation: Currently podio does not support building collections of "heterogeneous" types, i.e. it would not be easily possible to put measurements of different dimensions into a collection. Having different types for the different dimensions would be very much preferred for EDM4hep, mainly because an N dimensional "type erased" type, would make the first concern from above even more difficult to handle.

paulgessinger:
Hey @tmadlener, thanks for the input! 

I agree with your first point. Clearly this should be well-defined. However, I do think a local measurement + geometry connection is a reasonably well-defined concept that it doesn't allow "user-data", especially if we tie the axis interpretation to an enum that lives in edm4hep. This doesn't allow "user-data" any more than say the current `TrackerHitPlane` cov matrix.

[...]

PR #170, opened 2022-08-08
[wip] add `MeasurementND` classes

tmadlener:
BEGINRELEASENOTES
- TODO

ENDRELEASENOTES

This is a draft implementation for the `MeasurementND` classes mentioned in #161. In its current form it only implements a two dimensional example, but it should easily scale to other numbers of dimensions as well. It probably has some potential for generalizing the `ExtraCode` bits slightly for that.

In the current form it introduces a `Measurement2D` class that has three `std::array`s internally. One for the central values, one for the covariance matrix and one that stores "meta" information on the dimensions. The dimensions in this case are `enum`s (`enum class` also works). The meta information stores which dimension corresponds to which stored value and there is functionality to set/get the order of the dimensions. Using `ExtraCode` we essentially simply layer a second interface that grants access to the stored values via the dimensions directly. Potentially simpler shown than told:

```cpp
enum Cartesian : edm4hep::DimType { X = 0, Y, Z}; // internally we use uint16_t, DimType is just a typedef for that

auto m = MutableMeasurement2D();
// This measurement should describe one in a (Z, X) space
m.setDimensions(Cartesian::Z, Cartesian::X);
// Setting the value for Z
m.setValue(3.14f, Cartesian::Z);
// Setting a value in the covariance matrix (order of dimensions doesn't matter)
m.setCov(1.23f, Cartesian::X, Cartesian::Z);
```
Retrieving the values can be done similarly.

One thing that is not even attempted in the current version is to guard against access with a different dimension enum, e.g. using the above `m` the following works without problems
```cpp
enum class Polar { r, theta, phi };
// this DOES NOT have to be the same dimensions as above
const auto [dim1, dim2] = m.getDimensions<Polar>();
// dim1 == Polar::phi
// dim2 == Polar::r

auto val = m.getValue(Polar::phi);
// val == 3.14f
```

Since this interface is simply added to the one that is generated, the whole class is also usable without any specific enums.

@paulgessinger does this go into a direction that is useful for ACTS?

TODO:
- [ ] Geometry info / CellID field
- [ ] Error handling. What should happen if access with a dimension that is not stored is attempted?
- [ ] Other Ns
- [ ] Better naming of things?
- [ ] Release notes

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -418,6 +418,50 @@ datatypes :
       //TODO: edm4hep::Vertex  getEndVertex() { return  edm4hep::Vertex(  (getParticles(0).isAvailable() ? getParticles(0).getStartVertex() :  edm4hep::Vertex(0,0) ) ) ; }\n     
       "
 
+  edm4hep::Measurement2D:
+    Description: "Two dimensional measurement with covariance matrix"
+    Author: "T. Madlener, DESY"
+    Members:
+      - std::array<float, 2> vals    // The values of the measurement
+      - std::array<float, 3> covMatrix     // The (lower triangle) covariance matrix of the measurement
+      - std::array<uint16_t, 2> dims // The (encded) dimensions of the two values
+    ExtraCode:
+      includes: '#include "edm4hep/detail/enum_access.h"'
+      declaration: "
+      /// Get the value for the passed dimension\n
+      template<typename DimEnum>\n
+      float getValue(DimEnum e) const {\n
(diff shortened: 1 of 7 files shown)


paulgessinger:
I realize I never commented on this, sorry.
Thanks @tmadlener, at first glance, this looks useful for us. I haven't had time to look at it in more detail and test it out, but I plan to get back to this soon.

wdconinc:
I'm curious about this PR. Is this still of interest or is it being pursued elsewhere?

PR #172, opened 2022-08-17
Make sure that the kinematics.h also works for python

tmadlener:
BEGINRELEASENOTES
- Fix the `kinematics.h` header only utility file such that it also works when used via PyROOT.

ENDRELEASENOTES

For more details see:
- https://root-forum.cern.ch/t/calling-a-tag-dispatched-function-from-pyroot-does-not-work-with-a-default-tag-argument/51156/4
- wlav/cppyy#78


tmadlener:
Key4hep CI currently failing in the downstream tests due to not recent enough podio version. Merging this regardless.

PR #175, opened 2022-08-24
Move RDF utilities to utils folder and make utility functions templated

### LINKED ISSUE #102: dataframe utilities could be moved to utils directory
As decided during todays meeting, we now have a separate [EDM4Hep-utils](https://github.com/key4hep/EDM4hep-utils) repository. I think the contents of [dataframe](https://github.com/key4hep/EDM4hep/tree/master/dataframe) and the related tests could be moved there.

tmadlener:
BEGINRELEASENOTES
- Move the dataframe utilities to the utils folder
- Make the existing functionality templated and explicitly instantiate all possible datatypes
- Add a python RDataFrame test

ENDRELEASENOTES

The templated utility functions make the c++ RDataFrame slightly less convenient to use, because it cannot do overload resolution, so that we have to explicitly state the datatype for which we want to use a template function. Conversely for the python side the type is deduced from the branch we use the utility function on, so that here things are simpler.

Fixes #102


test/test_rdf.cc
```diff
@@ -1,27 +1,40 @@
...
   std::cout << "Create RDataFrame ..." << std::endl;
+
   ROOT::RDataFrame df("events", "edm4hep_events.root");
+
   std::cout << "Apply selectors and define new branches ..." << std::endl;
-  auto df2 = df.Define("MCParticles_pt", edm4hep::pt, {"MCParticles"})
-                 .Define("MCParticles_eta", edm4hep::eta, {"MCParticles"})
-                 .Define("MCParticles_cosTheta", edm4hep::cos_theta, {"MCParticles"})
-                 .Define("SimTrackerHits_r", edm4hep::r, {"SimTrackerHits"});
+  auto df2 = df.Define("MCParticles_pt", pt<edm4hep::MCParticleData>, {"MCParticles"})
```
clementhelsens, line 21: in the python example you explicitly add ```edm4hep::utils``` while in the cc example you hide it with using the namespace. As this is an example, I would suggest to explicitly add it here. Or alternatively, you can load the namespace in python ```ROOT.gInterpreter.Declare("using namespace edm4hep::utils;")``` such that both examples are consistent

clementhelsens: To show the template in the example, I would consider adding the same but for ```edm4hep::ReconstructedParticleData``` (same in python)

tmadlener: Very good points. I have removed the `using namespace edm4hep::utils` from the c++ example.

Unfortunately, there are no reconstructed particles in our test files yet, but I have added two more calls
- One to `edm4hep::utils::pt` for a `SimTrackerHit` (where also the momentum of the particle causing the hit is stored)
- One to `edm4hep::utils::r` for a `TrackerHitPlane`

Do you think that is enough. Alternatively I can also add some recos to the tests.

tmadlener: It would in principle also be possible to have the c++ and python examples *exactly* the same by also using JIT compilation there:
```cpp
gInterpreter->LoadFile("edm4hep/utils/dataframe.h"); // This is necessary to make JIT-ing work

ROOT::RDataFrame df("events", "edm4hep_events.root");

auto df2 = df.Define("MCParticles_pt", "edm4hep::utils::pt(MCParticles");
// ...
```

I am not entirely sure we should do that though. What do you think?

clementhelsens: Thanks @tmadlener , I think it's enough for the time being what you've added. But I suggest to keep in mind adding reconstructed particles later such that we also explicitly show relations once we have a ```Podio/RDataSource``` prototype from @forthommel

clementhelsens: > I am not entirely sure we should do that though. What do you think?

I think this is not connected to edm4hep and it is more a user preference. I would not enter this level of details here.

tmadlener: Adding more classes to the tests is on the list: #158

clementhelsens (approved):

test/test_rdf.py
```diff
@@ -0,0 +1,35 @@
...
+print('Cerate RDataFrame ...')
+df = ROOT.RDataFrame('events', 'edm4hep_events.root')
+
+print('Apply selectors and define new branches ...')
+df2 = (df.Define('MCParticles_pt', 'edm4hep::utils::pt(MCParticles)')
+       .Define('MCParticles_eta', 'edm4hep::utils::eta(MCParticles)')
+       .Define('MCParticles_cosTheta', 'edm4hep::utils::cos_theta(MCParticles)')
+       .Define('SimTrackerHits_r', 'edm4hep::utils::r(SimTrackerHits)')
+       .Define('SimTrackerHits_pt', 'edm4hep::utils::pt(SimTrackerHits)')
+
```
tmadlener, line 22:
```suggestion
```

PR #178, opened 2022-09-26
Updated line links to datatypes/components/associations

kjvbrt:
BEGINRELEASENOTES
- Updated line links to datatypes/components/associations in the README

ENDRELEASENOTES

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,29 @@ A generic event data model for future HEP collider experiments.
 
 **Componets**
 
+| | | |
+|-|-|-|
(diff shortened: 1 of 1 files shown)


tmadlener:
Looks good to me. Seems like CI currently breaks because a regex matches on some ROOT output (that should be harmless). The underlying issue will be fixed with AIDASoft/podio#334 .

tmadlener:
Running the workflows that build podio on the fly again to make sure the problem is actually fixed

Issue #25, opened 2020-02-04
`TPCHit` too specific?


vvolkl:
The content of [`TPCHit`](https://github.com/HSF/EDM4hep/blob/add_ReconstructedParticle/edm4hep.yaml#L283) is maybe useable for more detector types - could consider renaming.

tmadlener:
As per the discussion in the [EDM4hep meeting](https://indico.cern.ch/event/1204616/) (Oct. 18, 2022) this should be renamed to `TrackerRawData`.

@wenxingfang do you want to do this as part of #179, or should we rather make a separate pull request for it?

wenxingfang:
I can do it in https://github.com/key4hep/EDM4hep/pull/179. :)

PR #179, opened 2022-10-17
extend for drift chamber and TPC

wenxingfang:
BEGINRELEASENOTES
- Extend for drift chamber and TPC study
- The extended EDMs are tested within CEPCSW, and it works well
- The extended edm4hep::TrackerPulse and edm4hep::TrackerData are similar to what they are in ILC TPC

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -323,7 +323,7 @@ datatypes :
     Author : "F.Gaede, DESY"
     Members :
       - uint64_t cellID      //ID of the sensor that created this hit
-      - int32_t type                       //type of raw data hit, either one of edm4hep::TPCHIT, edm4hep::SIMTRACKERHIT - see collection parameters "TrackerHitTypeNames" and "TrackerHitTypeValues".
+      - int32_t type                       //type of raw data hit, either one of edm4hep::TrackerRawData, edm4hep::SIMTRACKERHIT - see collection parameters "TrackerHitTypeNames" and "TrackerHitTypeValues".
       - int32_t quality                    //quality bit flag of the hit.
       - float time                     //time of the hit.
       - float eDep                     //energy deposited on the hit [GeV].
@@ -340,7 +340,7 @@ datatypes :
     Author : "Placido Fernandez Declara, CERN"
     Members :
       - uint64_t cellID      //ID of the sensor that created this hit
-      - int32_t type                       //type of raw data hit, either one of edm4hep::TPCHIT, edm4hep::SIMTRACKERHIT - see collection parameters "TrackerHitTypeNames" and "TrackerHitTypeValues".
+      - int32_t type                       //type of raw data hit, either one of edm4hep::TrackerRawData, edm4hep::SIMTRACKERHIT - see collection parameters "TrackerHitTypeNames" and "TrackerHitTypeValues".
       - int32_t quality                    //quality bit flag of the hit.
       - float time                     //time of the hit.
       - float eDep                     //energy deposited on the hit [GeV].
@@ -355,8 +355,8 @@ datatypes :
       - edm4hep::ObjectID  rawHits     //raw data hits. Check getType to get actual data type.
 
 
- #----------  TPCHit
-  edm4hep::TPCHit:
+ #---------- TrackerRawData 
+  edm4hep::TrackerRawData:
     Description: "Time Projection Chamber Hit"
     Author : "F.Gaede, DESY"
     Members:
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -489,3 +489,75 @@ datatypes :
     OneToOneRelations:
      - edm4hep::ReconstructedParticle rec // reference to the reconstructed particle
      - edm4hep::Vertex vertex             // reference to the vertex
+
+
+  edm4hep::SimPrimaryIonizationCluster:
+    Description: "Simulated Primary Ionization"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+      - unsigned long long cellID            //cell id.
+      - float time                           //the primary ionization's time in the lab frame in [ns].
+      - edm4hep::Vector3d position           //the primary ionization's position in [mm].
+      - int   type                           //type.
+    VectorMembers:
+       - unsigned long long electronCellID   //cell id.
+       - float electronTime                  //the time in the lab frame in [ns].
+       - edm4hep::Vector3d electronPosition  //the position in the lab frame [mm].
+       - float pulseTime                     //the pulse's time in the lab frame in [ns].
+       - float pulseAmplitude                //the pulse's amplitude.
+    OneToOneRelations:
+      - edm4hep::MCParticle MCParticle       //the particle that caused the ionizing collisions.
+
+  edm4hep::TrackerPulse:
+    Description: "Reconstructed Tracker Pulse"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - unsigned long long cellID       //cell id. 
+       - float time                      //time.
+       - float charge                    //charge.
+       - int quality                     //quality.
+    VectorMembers:
+       - float covMatrix                 //covariance matrix of the charge(c) and time(t) measurements.
+    OneToOneRelations:
+      - edm4hep::TrackerData trackerData  //Optionaly, the TrackerData that has been used to create the pulse can be stored with the pulse.
+
+  edm4hep::RecIonizationCluster:
+    Description: "Reconstructed Ionization Cluster"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - unsigned long long cellID  //cell id. 
+       - float significance         //significance.
+       - int type                   //type.
+    OneToManyRelations:
+       - edm4hep::TrackerPulse trackerPulse //the TrackerPulse used to create the ionization cluster.
+
+  edm4hep::TrackerData:
+    Description: "TrackerData"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - unsigned long long cellID  //cell id. 
+       - float time                 //begin time.
+       - float interval             //interval of each sampling in [ns].
+    VectorMembers:
+       - float chargeValue          //charge value.
+
+
+  edm4hep::RecDndx:
+    Description : "dN/dx info of Track."
+    Author : "Wenxing Fang, IHEP"
+    Members :
+      - float dNdx                     //the reconstructed dNdx. 
+      - float dNdxError                //error on the dNdx.
+      - int particleType               //particle type, e(0),mu(1),pi(2),K(3),p(4).
+      - int type                       //type.
+    VectorMembers:
+      - unsigned long long  cellID     //cell id.
+      - float  N                       //number of reconstructed ionization cluster.
+      - float  edep                    //energy deposite.
+      - float  pathL                   //path length in [mm].
+      - float  chi                     //chi for e(0), mu(1), pi(2), K(3), p(4).
+      - float  dNdxExpect              //expected dNdx for e(0),mu(1), pi(2), K(3), p(4).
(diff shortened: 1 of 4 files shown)


edm4hep.yaml
```diff
@@ -489,3 +489,75 @@ datatypes :
...
+    Description : "dN/dx info of Track."
+    Author : "Wenxing Fang, IHEP"
+    Members :
+      - float dNdx                     //the reconstructed dNdx. 
+      - float dNdxError                //error on the dNdx.
+      - int particleType               //particle type, e(0),mu(1),pi(2),K(3),p(4).
+      - int type                       //type.
+    VectorMembers:
+      - unsigned long long  cellID     //cell id.
+      - float  N                       //number of reconstructed ionization cluster.
```
tmadlener, line 555:
```suggestion
      - int  N                       //number of reconstructed ionization cluster.
```
(as commented by Andre in the meeting)

tmadlener (commented):
See also discussion in [EDM4hep meeting](https://indico.cern.ch/event/1204616/)

I have one more technical comment, that I did not raise during the meeting. Could you make all integer types a fixed width integer, especially the `unsigned long long` are somewhat loosely defined with only a minimal size and might change width depending on compiler, platform, etc. We have done this exercise for all other datatypes in EDM4hep, so for consistency it would be nice if it can also be done here.

edm4hep.yaml
```diff
@@ -489,3 +489,75 @@ datatypes :
...
+  edm4hep::SimPrimaryIonizationCluster:
+    Description: "Simulated Primary Ionization"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+      - unsigned long long cellID            //cell id.
+      - float time                           //the primary ionization's time in the lab frame in [ns].
+      - edm4hep::Vector3d position           //the primary ionization's position in [mm].
+      - int   type                           //type.
+    VectorMembers:
+       - unsigned long long electronCellID   //cell id.
```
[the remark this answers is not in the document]
wenxingfang: as ionized electrons from the same primary ionization cluster may be produced in different drift chamber cells, especially when they are produced close to the edge of a cell. So each ionized electron should have one cell id.

edm4hep.yaml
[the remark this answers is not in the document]
wenxingfang: yes, you are right.

edm4hep.yaml
```diff
@@ -489,3 +517,74 @@ datatypes :
...
+
+  #----------- SimPrimaryIonizationCluster
+  edm4hep::SimPrimaryIonizationCluster:
+    Description: "Simulated Primary Ionization"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+      - uint64_t cellID                      //cell id.
+      - float time                           //the primary ionization's time in the lab frame in [ns].
+      - edm4hep::Vector3d position           //the primary ionization's position in [mm].
+      - int16_t type                         //type.
```
tmadlener, line 530: You change this from `int` to `int16_t`, however, `int` is usually 32 bits on 64 bit systems in c++. I am fine with either, just want to make sure that 16 bits is enough here.

edm4hep.yaml
```diff
@@ -107,6 +107,34 @@ components:
...
+        - float expect_e   // expect value for electron
+        - float expect_mu  // expect value for muon
+        - float expect_pi  // expect value for pion
+        - float expect_k   // expect value for kaon
+        - float expect_p   // expect value for proton
+        - float sigma_e    // sigma  value for electron
+        - float sigma_mu   // sigma  value for muon
+        - float sigma_pi   // sigma  value for pion
+        - float sigma_k    // sigma  value for kaon
+        - float sigma_p    // sigma  value for proton
```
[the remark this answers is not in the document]
wenxingfang: you are right, it is better!

wenxingfang (commented):
to solve issue #25

wenxingfang:
Hi @gaede , I renamed 'RecDndx' to 'RecDqdx' and used edm4hep::Quantity dQdx to save the type, value, and error. What is your opinion?

tmadlener:
It looks like there are still a few formatting and whitespace issues. Could you fix them as well?

- https://github.com/key4hep/EDM4hep/actions/runs/3683601506/jobs/6232363223#step:4:526 (order of headers)
- https://github.com/key4hep/EDM4hep/actions/runs/3683601506/jobs/6232363223#step:4:537 (order of headers)
- https://github.com/key4hep/EDM4hep/actions/runs/3683601506/jobs/6232363223#step:4:556 (looks like mainly trailing whitespaces here)

wenxingfang:
@tmadlener @andresailer, This time all the checks are successful. It seems at least 1 approving review is needed in order to merge the PR.

edm4hep.yaml
```diff
@@ -489,3 +505,72 @@ datatypes :
...
+      - edm4hep::MCParticle MCParticle       //the particle that caused the ionizing collisions.
+
+ #----------  TrackerPulse
+  edm4hep::TrackerPulse:
+    Description: "Reconstructed Tracker Pulse"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - uint64_t cellID                 //cell id.
+       - float time                      //time.
+       - float charge                    //charge.
```
andresailer, line 535: The charge is in coulomb or multiple or electron (or proton) charges?

edm4hep.yaml
```diff
@@ -489,3 +505,72 @@ datatypes :
...
+ #----------  TrackerData
+  edm4hep::TrackerData:
+    Description: "TrackerData"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - uint64_t cellID            //cell id.
+       - float time                 //begin time.
+       - float interval             //interval of each sampling in [ns].
+    VectorMembers:
+       - float chargeValue          //charge value.
```
andresailer, line 561: Unit?

edm4hep.yaml
```diff
@@ -107,6 +107,22 @@ components:
...
+        - float chi2       // chi2
+        - float expected   // expected value
+        - float sigma      // sigma value
+
+    # hit level data
+    edm4hep::HitLevelData:
+      Members:
+        - uint64_t cellID  // cell id
+        - uint32_t N       // number of reconstructed ionization cluster.
+        - float  eDep        // energy deposit.
```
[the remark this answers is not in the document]
andresailer: eDep is the reconstructed energy deposit?

edm4hep.yaml
```diff
@@ -339,8 +355,8 @@ datatypes :
       - edm4hep::ObjectID  rawHits     //raw data hits. Check getType to get actual data type.
 
 
- #----------  TPCHit
-  edm4hep::TPCHit:
+ #---------- TrackerRawData
+  edm4hep::TrackerRawData:
     Description: "Time Projection Chamber Hit"
```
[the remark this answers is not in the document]
wenxingfang: ok, how about using "Raw data of a tracker hit"?

edm4hep.yaml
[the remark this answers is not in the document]
wenxingfang: yes, it should be a reconstructed energy deposit.

edm4hep.yaml
wenxingfang: It is in coulomb.

edm4hep.yaml
wenxingfang: will add a unit with [fC]

wenxingfang:
@andresailer Thanks for your suggestions! The PR has been updated. Please let me know if you have any questions or suggestions.

edm4hep.yaml
```diff
@@ -489,3 +505,72 @@ datatypes :
...
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - uint64_t cellID            //cell id.
+       - float significance         //significance.
+       - int16_t type               //type.
+    OneToManyRelations:
+       - edm4hep::TrackerPulse trackerPulse //the TrackerPulse used to create the ionization cluster.
+
+ #----------  TrackerData
+  edm4hep::TrackerData:
```
andresailer, line 553:
```suggestion
  edm4hep::TimeSeries:
```

edm4hep.yaml
```diff
@@ -339,9 +355,9 @@ datatypes :
       - edm4hep::ObjectID  rawHits     //raw data hits. Check getType to get actual data type.
 
 
- #----------  TPCHit
-  edm4hep::TPCHit:
-    Description: "Time Projection Chamber Hit"
+ #---------- TrackerRawData
+  edm4hep::TrackerRawData:
```
andresailer, line 359:
```suggestion
  edm4hep::ADCTimeSeries:
```

andresailer:
```diff
- - int32_t rawDataWords //raw data (32-bit) word at i.
+ - int32_t adcCounts    //raw data (32-bit) word at i.

File: edm4hep.yaml [hunk 11]
```diff
@@ -339,9 +355,9 @@ datatypes :
...
+    Description: "Raw data of a tracker hit"
```
andresailer, line 360:
```suggestion
    Description: "Raw data of a detector read out"
```

edm4hep.yaml
```diff
@@ -489,3 +505,72 @@ datatypes :
...
+ #----------  TrackerData
+  edm4hep::TrackerData:
+    Description: "TrackerData"
+    Author : "Wenxing Fang, IHEP"
+    Members:
+       - uint64_t cellID            //cell id.
+       - float time                 //begin time [ns].
+       - float interval             //interval of each sampling [ns].
+    VectorMembers:
+       - float chargeValue          //charge value [fC].
```
andresailer, line 561:
```suggestion
       - float amplitude          // calibrated detector data.
```

edm4hep.yaml
```diff
@@ -489,3 +505,72 @@ datatypes :
...
+    Members:
+       - uint64_t cellID            //cell id.
+       - float significance         //significance.
+       - int16_t type               //type.
+    OneToManyRelations:
+       - edm4hep::TrackerPulse trackerPulse //the TrackerPulse used to create the ionization cluster.
+
+ #----------  TrackerData
+  edm4hep::TrackerData:
+    Description: "TrackerData"
```
andresailer, line 554:
```suggestion
    Description: "Calibrated Detector Data"
```

edm4hep.yaml
andresailer: Need an entry for interval?

tmadlener: I am wondering whether `ADCTimeSeries` is not making this too specific? (In this case it is one, but in other cases a more general concept might be better, instead of having to introduce another type later?)

wenxingfang: > I am wondering whether `ADCTimeSeries` is not making this too specific? (In this case it is one, but in other cases a more general concept might be better, instead of having to introduce another type later?)

How about using "RawTimeSeries"?

wenxingfang: > Need an entry for interval?

I think it is needed.

wenxingfang:
@andresailer @tmadlener, thanks for the comments!  The PR has been updated. Welcome with further suggestions!

tmadlener (commented):
Thanks for all the updates and the patience. This looks good to me now.

It would be nice if we can address the necessary README changes still in this PR, but since we have to update the diagram as well, I would also be OK with doing that in a separate PR, unless @andresailer has a strong opinion on that.

For the repositories that break with these changes mentioned by @jmcarcell, I suppose most of the changes should be rather straight forward fixes, and we could prepare those PRs before we merge this and then do a coherent build for all of them to see if we didn't miss anything.

tmadlener:
There still seem to be some conflicts, although github doesn't really tell us where. I will try to have a look at this locally

tmadlener:
nevermind, the conflict is only for rebasing. merging should be fine

wenxingfang:
@jmcarcell Thanks a lot! Who should press the green button?

jmcarcell:
It's done now @wenxingfang

PR #185, opened 2022-11-15
Make building of json support dependent on nlohmann_json version

tmadlener:
BEGINRELEASENOTES
- Only build JSON support if a suitable version of nlohmann/json is found. This should fix a few issues in CI.

ENDRELEASENOTES

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,20 +1,22 @@
 list( APPEND CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" )
 
-add_executable(edm4hep2json src/edm4hep2json.cxx)
-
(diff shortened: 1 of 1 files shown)


tmadlener:
key4hep nightlies are failing because of https://github.com/HEP-FCC/FCCAnalyses/issues/227

vvolkl:
Thanks for the fix - something is still wrong though when fccanalyses can break the edm4hep tests, that's not how dependencies should work... Maybe  a name collision in the rfd test?

tmadlener:
Yeah, I agree that this really should not happen, but I suspect the culprit in this case is ROOT that loads dictionaries a bit too eagerly in this case. I suppose it would be possible to set the test environment even stricter to exclude, e.g. other dictionaries to be discoverable during tests. However, I seem to remember, that I have struggled with this already quite a bit in the past, due to how different e.g. an LCG (view) environment is compared to one from a key4hep stack build.

PR #188, opened 2023-01-09
Change in edm4hep.yaml for consistent naming

FinnJohannsen:
BEGINRELEASENOTES
- Changed the name of one VectorMember of edm4hep::track from subDetectorHitNumbers to subdetectorHitNumbers to be consistent with other spellings of the word subdetector in the yaml file and in LCIO.

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -381,7 +381,7 @@ datatypes :
       - float dEdxError                  //error of dEdx.
       - float radiusOfInnermostHit       //radius of the innermost hit that has been used in the track fit
     VectorMembers:
-      - int32_t subDetectorHitNumbers        //number of hits in particular subdetectors.Check/set collection variable TrackSubdetectorNames for decoding the indices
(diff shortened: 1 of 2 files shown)


tmadlener:
Yes, this definitely affects other repos as well, plus it also requires schema evolution support from podio. So for now, we will just leave this open, and then go for a concerted effort with all other repositories affected.

tmadlener:
To avoid too many problems in dependencies, we need to tag EDM4hep pretty much immediately after this has been merged.

tmadlener:
In a local build of the key4hep stack this breaks a few more things than just the converter. I will prepare PRs to fix those as well.

tmadlener:
The PRs in CEPCSW and k4LCIOReader should make this work smoothly once they are merged.

jmcarcell:
Why the `CMakeLists.txt` changes though? Those are unrelated and not needed...

tmadlener:
I think they are necessary in some of our CI workflows for a cleaner test environment (or at least one that has the "underlying" environment late enough in order to not pick it up). I must confess, I overlooked these when merging, but I don't think they harm us either.

PR #193, opened 2023-02-21
Add a configuration file for the podio visualization tool

jmcarcell:
BEGINRELEASENOTES
- Add a configuration file for the new podio visualization tool

ENDRELEASENOTES

To reproduce the last diagram shown in https://github.com/AIDASoft/podio/pull/377 (i.e. with clustering the data types into categories) it is necessary to have a configuration file that defines which groups exists. While it can be in either place I believe it should be here so changes in the data types such as renames are more likely to be applied also to this configuration file.

As a reminder, to get the diagram run (once it's merged and installed) `python path_to_podio/tools/podio-vis path_to_EDM4hep/edm4hep.yaml --graph-conf path_to_EDM4hep/utils/graph-conf.yaml`

We can also agree on which Associations to remove to make the plot nicer which can be removed by adding

``` yaml
Filter:
  - datatype
```


jmcarcell:
This is how it looks like with the current configuration file without removing any association

diagram

PR #201, opened 2023-03-31
Include sstream header explicitly

paulgessinger:
The header file uses it, but it's not included explicitly (and actually
fails to compile locally for me because of this).

BEGINRELEASENOTES
- Added string stream include in `edm4hep2json.hxx`

ENDRELEASENOTES

### DIFF
--- a/tools/include/edm4hep2json.hxx
+++ b/tools/include/edm4hep2json.hxx
@@ -28,6 +28,7 @@
 #include <exception>
 #include <fstream>
 #include <iostream>
+#include <sstream>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


jmcarcell (approved):
No objections. I think I know why it is failing for you but not for the workflows or for me. `sstream` is being currently included through one of the podio includes but if you are using a podio from before 2023-03-06 then that file doesn't have `#include <sstream>`. Had to dig a bit since it would be weird that no one had complained before it this was broken using the master branches.

tmadlener (approved):

PR #204, opened 2023-04-22
Add python bindings for the datamodel classes

jmcarcell:
BEGINRELEASENOTES
- Add python bindings for the datamodel classes and some documentation on how to use the bindings

ENDRELEASENOTES

See the README for examples on how to use them


README.md
```diff
@@ -72,10 +72,24 @@ make test
 
 The library files and dictionaries (`libedm4hep.so`, ...) must be put somewhere in `LD_LIBRARY_PATH`.
 
-
-
-
-
+## Python bindings
```
tmadlener, line 75: Maybe clarify here that this is only necessary when users are not reading stuff via the podio python bindings. Because they will take care of everything for only working with collections read form file.

python/edm4hep/datamodel.py
```diff
@@ -0,0 +1,14 @@
+import importlib
```
tmadlener, line 1: Can the name of this file start with a `_` to indicate that it is "private"?

tmadlener (commented):
Thanks for taking care of this. This already looks pretty good, My main complaint is the factory being something that the users can see. I suppose we need the factory because of the dynamic nature of ROOTs python bindings, right? Can we still try to "hide" it in the `__init__.py` or in `datamodel.py` such that it is already instantiated and what the user sees would look something like:

```python
from edm4hep import edm4hep
particle = edm4hep.MCParticle()
# ...
```
This should also work with `from edm4hep import *`, but I am not sure I like that better. Ideally it would just be `import edm4hep` and then we have the factory in hand, but I don't see a way how that could work.

python/edm4hep/datamodel.py
```diff
@@ -0,0 +1,14 @@
...
+import ROOT
+ROOT.gSystem.Load('libedm4hepDict.so')
```
tmadlener, line 3: This returns either `0`, `1` or `-1`, where `-1` indicates a failure to load, that could be used to raise an `ImportError` with a message. If the load succeeded, `from ROOT import edm4hep` should work without problems and we wouldn't need the `importlib`

jmcarcell: Fixed but `importlib` is needed because even though you can do `from ROOT import edm4hep` the names of the classes don't exist until they are called explicitly so either you load all of them at the beginning or you load them dynamically like it's done here:
https://github.com/key4hep/EDM4hep/blob/python-bindings/python/edm4hep/datamodel.py#L17

README.md
jmcarcell: Done

python/edm4hep/datamodel.py
jmcarcell: Actually no, it's possible to do it and `importlib` has been removed

python/edm4hep/datamodel.py
jmcarcell: I'd rather not, I don't think that convention is used in modern python code

jmcarcell:
It has been changed now to `edm4hep` so both work:
```
>>> from edm4hep import *
>>> edm4hep
<edm4hep.datamodel.Datamodel object at 0x7f4bd08fa0b0>
>>> edm4hep.MCParticle()
<cppyy.gbl.edm4hep.MCParticle object at 0x55b3f52c4710>
```
or the one in the README:
```
>>> from edm4hep import edm4hep
>>> edm4hep.MCParticle()
<cppyy.gbl.edm4hep.MCParticle object at 0x55b4034ba030>
```
`import edm4hep` doesn't work since it's the top level module... And maybe it's not a good idea in case we have other python tools in the future for edm4hep

tmadlener:
Thanks this looks very good now. I just realized we do have some tests for utility functionality already, that redo some of the stuff that is introduced here:

https://github.com/key4hep/EDM4hep/blob/1145e20b9ee238e1110f643232cabd061646582e/test/utils/test_kinematics.py#L10-L16

Now that we have a proper place for an `__init__.py` maybe we could even use that to also include the JIT  loading of the `utils/kinematics.h` header properly instead of the hackish way we do it at the moment for the tests.

tmadlener:
Even more cases in the tests that could be cleaned up:
https://github.com/key4hep/EDM4hep/blob/1145e20b9ee238e1110f643232cabd061646582e/test/test_rdf.py#L5-L9

jmcarcell:
That is done now. I think it could be done on demand when using python but in the cases where something in `edm4hep::utils` is passed as a string python has no way to know so this will always load both headers. For running the tests manually it's a bit annoying that one has to set the paths but it's hard to make it possible to load the headers always. I tried and had a working version for tests and runtime without needing to set the paths but then someone could break it by installing instead of in `install/` in the main tree in `install1/install2` or making an in-source build

README.md
```diff
@@ -72,12 +72,27 @@ make test
...
+There are python bindings for all the classes in the datamodel for working with
+the clases individually (for working with collections podio has its own
+bindings). Make sure that `install/lib` is in `LD_LIBRARY_PATH` and
+`install/python` is in `PYTHONPATH`:
```
hegner, line 79: can you replace `install` with the proper CMake variable in there? just to be very explicit (and then you can add `(e.g. install/python`),

python/edm4hep/datamodel.py
```diff
@@ -0,0 +1,29 @@
+import ROOT
+res = ROOT.gSystem.Load('libedm4hepDict.so')
+if res < 0:
+    raise RuntimeError('Failed to load libedm4hepDict.so')
+
+kinematics_path = 'edm4hep/utils/kinematics.h'
```
hegner, line 6: kinematics is something used for the tests or provided as default?

hegner (changes requested):

python/edm4hep/datamodel.py
tmadlener: It's a utility header that we ship by default.

hegner: thanks

README.md
jmcarcell: Fixed

python/edm4hep/datamodel.py
```diff
@@ -0,0 +1,30 @@
...
+if res < 0:
+    raise RuntimeError('Failed to load libedm4hepDict.so')
+
+res = ROOT.gSystem.Load('libedm4hepRDF.so')
+if res < 0:
+    raise RuntimeError('Failed to load libedm4hepRDF.so')
+
+res = ROOT.gInterpreter.LoadFile('edm4hep/utils/kinematics.h')
+if res != 0:
+    raise RuntimeError('Failed to load kinematics.h')
```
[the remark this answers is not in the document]
jmcarcell: I made some changes in https://github.com/key4hep/EDM4hep/pull/204/commits/81d328bfec966c6cfa4b311ad22cd3c41df9594c and now it's possible:
``` python
from edm4hep import edm4hep
particle = edm4hep.MCParticle() # default initialized particle
particle.getCharge() # 0.0

from edm4hep import utils
p  = utils.p4(particle) #edm4hep.utils.p4 or utils.p4 works
print(p) # (0,0,0,0)
```
There are two classes now, one for serving all classes from edm4hep and another one for the utils namespace. If there were more namespaces it's easy to change to a dictionary-based class that would store each of the `edm4hep.namespace` object and serve by doing a lookup on the dictionary

python/edm4hep/datamodel.py
```diff
@@ -0,0 +1,43 @@
+import ROOT
+res = ROOT.gSystem.Load('libedm4hepDict.so')
+if res < 0:
+    raise RuntimeError('Failed to load libedm4hepDict.so')
```
tmadlener, line 4: I am wondering whether all these `RuntimeError`s should be `ImportError`s instead. In the end `import edm4hep` is the thing that will fail if one of these is triggered.

tmadlener (commented):
Very nice, thanks a lot for having another look at this. Since we are already at it, I think we could also add a `__version__` to the `__init__.py` for an even more pythonic feel.

In podio we have a `__init__.py.in` that we configure during the cmake stage. This should be trivial to add here as well, I think.
- [`configure_file` in cmake](https://github.com/AIDASoft/podio/blob/930e6001e893bef9415639710e12b67b5deba5d6/CMakeLists.txt#L164C10-L165)
- [`__init__.py.in`](https://github.com/AIDASoft/podio/blob/master/python/__init__.py.in)

python/edm4hep/datamodel.py
jmcarcell: Well typically for the libraries it will be because they are not in `LD_LIBRARY_PATH` which is a runtime search so one could say it's a `RuntimeError` ... both are quite generic, I don't think `ImportError` tells you much, that will also be quite obveious when checking the trace. So I don't have a strong opinion on this.

jmcarcell:
I added a version file. I preferred that over having an `__init__.py.in` so that it's possible to have the module loaded when developing and have suggestions, fixes, etc right away:
2023-06-14-152148_1254x156_scrot
and if the version is not there it will complain but keep loading everything else.

python/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+install(DIRECTORY edm4hep DESTINATION python)
```
tmadlener, line 1:
```suggestion
install(DIRECTORY edm4hep DESTINATION python
  REGEX .*\\.in EXCLUDE   # don't install configuration input files
  PATTERN __pycache__ EXCLUDE  # don't install python caches
  )
```
Final comment to make sure to only install what we want

tmadlener (commented):
Really final comment (:wink:): I think it would make sense to add `python/edm4hep/__version__.py` to `.gitignore`

python/edm4hep/datamodel.py
tmadlener: Yeah, maybe a `RuntimeError` is better, because that doesn't point you towards a python problem in this case.

tmadlener (approved):

PR #206, opened 2023-06-28
Update schema diagram

tmadlener:
BEGINRELEASENOTES
- Replace TPCHit with RawTimeSeries
- Update version to 0.9
- Add TrackerPulse

ENDRELEASENOTES

### DIFF
--- a/doc/edm4hep_diagram.svg
+++ b/doc/edm4hep_diagram.svg
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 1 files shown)


jmcarcell (commented):
Looks good although I don't like how the boxes are smaller now when there is plenty of space but I wouldn't bother to change it.
For the new types there is a one to one relation between each RawTimeSeries and each TrackerPulse but I don't see an arrow. I don't even know if all the relations and associations between classes present in the diagram are drawn so I can't really say it should be there.

tmadlener:
Relations are not trivial to draw in this case, because the `TrackerPulse` has a relation to a `TimeSeries` (but not to the `RawTimeSeries`). Additionally, I haven't put all the types for the drift chamber study on there yet, because I am not entirely sure where to place them. I could also just bump the version and remove the `TrackerPulse` again, which would bring us to the old diagram (with `TPCHit` -> `RawTimeSeries`).

The boxes where changed automatically when opening the file.

jmcarcell:
I think it's OK

PR #207, opened 2023-06-30
Improve python API, use `import edm4hep` instead of `from edm4hep import edm4hep`

jmcarcell:
BEGINRELEASENOTES
- Improve python API, use `import edm4hep` instead of `from edm4hep import edm4hep`
ENDRELEASENOTES

This simplifies the code internally and also allows us to have TAB completion in an interactive python session. I prefer `import edm4hep` to `from edm4hep import edm4hep`; I think it makes more sense that you get EDM4hep and its classes directly, and it's the python way of importing a module for almost every module. Then if we add any additional python feature it can be added and imported with `from edm4hep import feature`. This breaks `from edm4hep import edm4hep`, which could be preserved by adding

``` python
sys.modules['edm4hep.edm4hep'] = edm4hep
```


tmadlener (approved):
:star_struck: Now this is awesome. Thanks.

PR #210, opened 2023-07-04
Add /.cache/ to gitignore (tooling section)

tmadlener:
BEGINRELEASENOTES
- Update `.gitignore` to not pick up on newer directories of `clangd`

ENDRELEASENOTES

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ spack*
 /.clangd/
 /compile_commands.json
 *__pycache__*
+/.cache/


jmcarcell:
I have nothing against this but I think we should have this be managed at once for all the key4hep repos instead of adding it one by one, let me see if I can have a look today

tmadlener:
If that would be possible I would very much prefer that

andresailer:
@tmadlener: Are you making use of the global git-ignore? https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

tmadlener:
Not yet actually. I should probably. Maybe we could also provide a sort of default one for Key4hep as a starting point for people?

jmcarcell:
Yes, first we should agree on a default one and then I have a tool I can adapt to push the changes to most of the repos. See https://github.com/key4hep/key4hep-dev-utils/pull/1 for an example

jmcarcell:
After https://github.com/key4hep/key4hep-dev-utils/pull/1 a default `.gitignore` was added and pushed to EDM4hep among others so this is no longer relevant unless there are more paths to add

PR #211, opened 2023-07-06
Add a OneToMany relation from TrackerHit to RecIonizationCluster

BrieucF:
With the objects added in [PR#179](https://github.com/key4hep/EDM4hep/pull/179) for cluster counting algorithms, the only way to get the number of ionizations associated to a given `TrackerHit` is to do a matching by `cellID` which is quite inefficient. This PR proposes to add a link between `TrackerHit` and `RecIonizationCluster` to simplify the derivation of dN/dx.

BEGINRELEASENOTES
- Add a OneToMany relation from TrackerHit to RecIonizationCluster to allow us deriving the number of clusters associated to a TrackerHit without having to do a matching by cellID

ENDRELEASENOTES


BrieucF:
I would update the line numbers in the `README.md` once we agree on the way to go, before merging. By the way, isn't there a script somewhere that does it automatically?

tmadlener:
> By the way, isn't there a script somewhere that does it automatically?

No, but there really should be ;)

andresailer:
Closing (at least temporarily) while DriftChamber Tracker hit is under development by @BrieucF

PR #212, opened 2023-07-07
Add a script to automatically update the README.md links

BrieucF:
BEGINRELEASENOTES
- Add a script to automatically update the README.md links

ENDRELEASENOTES

### DIFF
--- a/scripts/updateReadmeLinks.py
+++ b/scripts/updateReadmeLinks.py
@@ -1,0 +1,50 @@
+#Usage: python3 scripts/updateReadmeLinks.py
+import re
+import sys
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("inputfile", help="The README file to run on", default="README.md")
+parser.add_argument("--check", action='store_true', help='Use if you want to just check the links without actually changing the file')
+args = parser.parse_args()
+
+def check_readme_links():
+  edm4hep_yaml_content = ""
+  with open("edm4hep.yaml") as edm4hep_yaml:
+    edm4hep_yaml_content = edm4hep_yaml.readlines()
+
+  readme_content = ""
+  with open("README.md", "r") as readme:
+    readme_content = readme.readlines()
+
+  new_readme_content = ""
+
+  links_are_ok = True
+  for readme_line in readme_content:
+    edm4hep_objects = re.findall("\[(.*?)\]\(https:\/\/github.com\/key4hep\/EDM4hep\/blob\/master\/edm4hep\.yaml#L(\d+?)\)", readme_line)
+    if edm4hep_objects:
+      for edm4hep_object, original_line_number in edm4hep_objects:
+        regex = f".*edm4hep::{edm4hep_object} *: *"
+        # find in which line it appears in edm4hep.yaml
+        edm4hep_yaml_line_numbers_with_match = [str(i + 1) for i, line in enumerate(edm4hep_yaml_content) if re.match(regex, line)]
+        if len(edm4hep_yaml_line_numbers_with_match) != 1:
+          print(f"Error: failed to replace line number for {edm4hep_object}, either no or several matches were found in edm4hep.yaml with the regex '{regex}'")
+          sys.exit(1)
+        if edm4hep_yaml_line_numbers_with_match[0] != original_line_number:
+          links_are_ok = False
+          readme_line = readme_line.replace(f"[{edm4hep_object}](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L{original_line_number})", f"[{edm4hep_object}](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L{edm4hep_yaml_line_numbers_with_match[0]})")
+          print(f"{edm4hep_object} is wrongly linked (line {original_line_number} --> {edm4hep_yaml_line_numbers_with_match[0]})")
+    new_readme_content += readme_line
+  return links_are_ok, new_readme_content
+
+links_are_ok, new_readme_content = check_readme_links()
+if links_are_ok:
+  print("README.md links are fine, nothing to change")
+else:
+  if not args.check:
+    with open("README.md", "w") as readme:
+      readme.write(new_readme_content)
+    print("README.md links updated.")
+  else:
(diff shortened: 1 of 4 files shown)


scripts/updateReadmeLineNumbers.py
```diff
@@ -0,0 +1,37 @@
...
+  readme_content = readme.readlines()
+
+new_readme_content = ""
+
+for readme_line in readme_content:
+  edm4hep_objects = re.findall("\[(.*?)\]\(https:\/\/github.com\/key4hep\/EDM4hep\/blob\/master\/edm4hep\.yaml#L(\d+?)\)", readme_line)
+  if edm4hep_objects:
+    for object_lineNumber_tuple in edm4hep_objects:
+      edm4hep_object = object_lineNumber_tuple[0]
+      original_line_number = object_lineNumber_tuple[1]
```
jmcarcell, line 20:
```suggestion
    for edm4hep_object, original_line_number in edm4hep_objects:
```

scripts/updateReadmeLineNumbers.py
```diff
@@ -0,0 +1,37 @@
...
+      regex = ".*edm4hep::%s:$"%edm4hep_object
```
jmcarcell, line 21: This doesn't work if someone adds spaces after the colon
```suggestion
      regex = ".*edm4hep::%s: *$"%edm4hep_object
```

scripts/updateReadmeLineNumbers.py
```diff
@@ -0,0 +1,37 @@
...
+      edm4hep_yaml_line_number_with_match = -1
+      for idx, lineString in enumerate(edm4hep_yaml_content):
+        if re.match(regex, lineString):
+          edm4hep_yaml_line_number_with_match = str(idx + 1)
+          nmatch += 1
+      if nmatch != 1:
+        print("Error: failed to replace line number for %s, either no or several matches were found in edm4hep.yaml with the regex '%s'"%(edm4hep_object, regex))
+        sys.exit(1)
+      readme_line = readme_line.replace("[%s](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L%s)"%(edm4hep_object, original_line_number), "[%s](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L%s)"%(edm4hep_object, edm4hep_yaml_line_number_with_match))
+      print("Will replace line number from %s to %s for the object %s"%(original_line_number, edm4hep_yaml_line_number_with_match, edm4hep_object))
```
jmcarcell, line 33:
```suggestion
      matches = [i + 1 for i, line in enumerate(edm4hep_yaml_content) if re.match(regex, line)]
      if len(matches) != 1:
        print("Error: failed to replace line number for %s, either no or several matches were found in edm4hep.yaml with the regex '%s'"%(edm4hep_object, regex))
        sys.exit(1)
      readme_line = readme_line.replace("[%s](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L%s)"%(edm4hep_object, original_line_number), "[%s](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L%s)"%(edm4hep_object, matches[0]))
      print("Will replace line number from %s to %s for the object %s"%(original_line_number, matches[0], edm4hep_object))
```
More pythonic

jmcarcell (changes requested):
Looks good. Will people remember to use it? We'll see...
Can we not change edm4hep.yaml to remove those spaces? With the changes I propose that is not needed. They will make git blame a bit less useful since many lines will point to this commit and you can easily fix it by pasting the current version of edm4hep.yaml. 
If we want the changes it's fine but then let them be in a different commit that we can add to gitignore revs.

tmadlener:
Very nice. Thanks for picking this up (that was not my intention with the comment in #211). How hard would it be to extract a function that simply checks whether the lines in the yaml match with the links in the README? That would be rather easy to hook up to CI, I think and then at least we would have a check so that people would know to run the script.

Style(?) comment: We could use f-string formatting since this will only ever be run with python3, but that is more of a personal preference.

tmadlener:
For the whitespace changes in the yaml file, I would say we change the ones that are directly in the yaml structure, e.g. [these](https://github.com/key4hep/EDM4hep/pull/212/files#diff-41291f568217a9755476ae21a12825127ae5a00179b19e867cffd64e3ceea36fL11-R11), but we do not touch the ones that are in `ExtraCode` blocks.

BrieucF:
> How hard would it be to extract a function that simply checks whether the lines in the yaml match with the links in the README? That would be rather easy to hook up to CI, I think and then at least we would have a check so that people would know to run the script.

Would this work for the CI?

scripts/updateReadmeLinks.py
```diff
@@ -0,0 +1,40 @@
...
+    new_readme_content += readme_line
+  return links_are_ok, new_readme_content
+
+links_are_ok, new_readme_content = check_readme_links()
+if links_are_ok:
+  print("README.md links are fine, nothing to change")
+else:
+  with open("README.md", "w") as readme:
+    readme.write(new_readme_content)
+  print("README.md links updated.")
```
tmadlener, line 40: Can we add a `--check` flag to the script that does not write a new file and exits with a non-zero exit code if the links are not OK? Then it would be trivial to write a github action around it, I think. Potentially we also need the `README.md` as an input argument for CI, but we could check first.

BrieucF: Can you try with this?

tmadlener: Very nice. I think I can even integrate this as a simple `pre-commit` hook. Let me give it go quickly.

tmadlener:
```console
$ pre-commit run -a readme-links-check
README links check.......................................................Failed
- hook id: readme-links-check
- exit code: 1

Vector3f is wrongly linked (line 23 --> 11)
README.md links should be updated. (Use the scripts/updateReadmeLinks.py script)
```

This seems to work as intended if I artificially change a line number.

jmcarcell:
Not much, other than if this is squashed then the commits that change whitespaces and the changes will be in the same commit. It's actually well separated in two commits so removing those 2 should be use and they can go in another PR

tmadlener:
Should be done now.

tmadlener:
Thanks again @BrieucF for putting this script together.

PR #213, opened 2023-07-13
Rename CMAKE_{SOURCE,BIN}_DIR -> PROJECT_{SOURCE,BIN}_DIR

jmcarcell:
BEGINRELEASENOTES
- Rename `CMAKE_{SOURCE,BIN}_DIR` -> `PROJECT_{SOURCE,BIN}_DIR`

ENDRELEASENOTES

Makes it possible to build using add_subdirectory. I don't think there are any downsides.


tmadlener (approved):
This seems to be the proper thing to do. So let's do it.

jmcarcell:
This can also be done in more repos, I'll get to those later

PR #216, opened 2023-07-15
Make aliases global

jmcarcell:
BEGINRELEASENOTES
-  Make cmake aliases global to allow the aliases to be seen by other subdirectories at configure time
-  Change cmake aliases used within EDM4hep to `EDM4HEP::` instead of `edm4hep::` since the namespace for external use is `EDM4HEP`, for consistency

ENDRELEASENOTES

### DIFF
--- a/edm4hep/CMakeLists.txt
+++ b/edm4hep/CMakeLists.txt
@@ -4,6 +4,8 @@
 PODIO_GENERATE_DATAMODEL(edm4hep ../edm4hep.yaml headers sources IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS})
 
 PODIO_ADD_DATAMODEL_CORE_LIB(edm4hep "${headers}" "${sources}")
+add_library(edm4hep::edm4hep ALIAS edm4hep)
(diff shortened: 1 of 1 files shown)


andresailer:
The release notes would benefit from clarifying that these changes are related to cmake (because if I want to look for changes related to cmake I would search for cmake, for example, and "alias" could refer so many other things.). And also that the `EDM4HEP::` namespace for the targets is for internal use only, as the external `EDM4HEP::` is not affected at all.
https://github.com/key4hep/EDM4hep/blob/678f649857e3f505345a2df24349f18605b3a2da/cmake/EDM4HEPCreateConfig.cmake#L28

And we shouldn't confuse downstream users thinking they would need to change their use of EDM4HEP in the cmake.

You could even include the phrase outside the release notes into the release notes to tell people why this change happened.

jmcarcell:
I just edited the release notes and added more explanations. Indeed this is only for internal usage, nothing should change for other repositories.

jmcarcell:
Good to merge if it doesn't bother anyone I guess. It shouldn't change anything for the current builds; the downstream build is failing due to k4Pandora but that's unrelated to this...

PR #219, opened 2023-07-17
Respect CMAKE_INSTALL_BINDIR

jmcarcell:
BEGINRELEASENOTES
- Respect CMAKE_INSTALL_BINDIR

ENDRELEASENOTES

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -16,6 +16,6 @@ if (nlohmann_json_FOUND)
                                             edm4hep
                                             nlohmann_json::nlohmann_json)
 
-  install(TARGETS edm4hep2json DESTINATION bin)
+  install(TARGETS edm4hep2json DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 endif()


CMakeLists.txt
```diff
@@ -23,6 +23,11 @@ find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics)
 # - Use GNU-style hierarchy for installing build products
 include(GNUInstallDirs)
 
+# Build to lib instead of lib64:
+if (CMAKE_INSTALL_LIBDIR STREQUAL lib64)
+  set(CMAKE_INSTALL_LIBDIR lib)
+endif()
```
tmadlener, line 29: These seems to go exactly against what `GNUInstallDirs` does?

jmcarcell: Well yes it would overwrite `CMAKE_INSTALL_LIBDIR` but I think it's worth it to do since sometimes lib64 is the default (Alma 9 builds...). This would be bad for someone wanting to install to lib64 but that's a very rare use case? All builds are 64 bit. I was thinking about this yesterday but I'm not 100% sure what to do to solve the lib / lib64 mess; having that separation has only caused problems with 0 benefits

tmadlener: Yeah, it's annoying that Ubuntu and anything *centOS-like* do different things here. Not sure what the best solution here is either. However, we have `GNUInstallDirs` in (almost?) all of the key4hep packages, I think. That would mean we would have to roll out these changes to all of them.

jmcarcell: Ok I just dropped that part, we can pass it to spack maybe instead of changing it in every repo. What we can change in the repos is making sure the `CMAKE_INSTALL_<dir>` variables are respected which is not always the case but we don't see that in our current builds since it's a default path.

PR #221, opened 2023-08-24
prepare edm4hep for reflex to rootcling migration of PODIO

hegner:
BEGINRELEASENOTES
- Move edm4hep to use rootcling instead of reflex for dictionaries (following PODIO)
ENDRELEASENOTES

This change depends on https://github.com/AIDASoft/podio/pull/460

### DIFF
--- a/python/edm4hep/__init__.py
+++ b/python/edm4hep/__init__.py
@@ -5,7 +5,7 @@ from .__version__ import __version__
 import ROOT
 res = ROOT.gSystem.Load('libedm4hep.so')
 if res < 0:
-    raise RuntimeError('Failed to load libedm4hepDict.so')
+    raise RuntimeError('Failed to load libedm4hep.so')
 
 res = ROOT.gSystem.Load('libedm4hepRDF.so')
 if res < 0:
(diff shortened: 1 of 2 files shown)


tmadlener:
With this I can build EDM4hep on top of AIDASoft/podio#460 .

Looking at the bigger picture, I am wondering if we have / want to keep the cmake macro for the dictionary generation, since we no longer have a core (I/O free) library and the dictionaries. At the moment that effectively serves as a toggle to add the ROOT I/O to now single shared library.

python/edm4hep/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 from .__version__ import __version__
 import ROOT
-res = ROOT.gSystem.Load('libedm4hepDict.so')
+res = ROOT.gSystem.Load('libedm4hep.so')
 if res < 0:
     raise RuntimeError('Failed to load libedm4hepDict.so')
```
jmcarcell, line 8:
```suggestion
    raise RuntimeError('Failed to load libedm4hep.so')
```

tmadlener:
This is no longer relevant as podio follows a different approach now.

PR #223, opened 2023-09-04
CI: use same lcg stacks as podio

andresailer:
BEGINRELEASENOTES
- CI: use same lcg stacks as podio
- Test: update to Catch2 3.4.0, same as in podio, and c++20 compatible

ENDRELEASENOTES


tmadlener (approved):
Thanks.

PR #224, opened 2023-09-08
Skip catch test discovery if environment requires it

tmadlener:
BEGINRELEASENOTES
- Add `SKIP_CATCH_DISCOVERY` option to turn of `catch_discover_tests` which may not run in the right environment in older cmake versions.

ENDRELEASENOTES

See https://github.com/AIDASoft/podio/pull/425 for similar changes in podio.


tmadlener:
Merging this to see if it indeed fixes the CI issue in https://github.com/AIDASoft/podio/pull/477

PR #227, opened 2023-09-20
edm4hep2json: All EDM4hep collections + ROOT legacy reader

kjvbrt:
BEGINRELEASENOTES
- edm4hep2json now converts all EDM4hep collections, associations and Podio user data
- added ROOT legacy support for edm4hep2json

ENDRELEASENOTES

### DIFF
--- a/test/tools/test_all_collections.py
+++ b/test/tools/test_all_collections.py
@@ -1,8 +1,9 @@
 '''
-Tests if all collection header files are included in the cxx file.
+Tests if all datatypes are used in the cxx file.
 '''
 
 import sys
+import re
 import argparse
 import yaml
 
@@ -16,23 +17,27 @@ def test(yamlfile_path, cxxfile_path):
     with open(yamlfile_path, mode='r', encoding="utf-8") as yamlfile:
         datamodel = yaml.safe_load(yamlfile)
 
+    # List stores lines of cxx code on which `insertToJson<CollType>` is used
+    datatypes_found = []
+
     with open(cxxfile_path, mode='r', encoding="utf-8") as cxxfile:
-        cxxfile_lines = cxxfile.readlines()
+        for cxxline in cxxfile:
+            cxxline = cxxfile.readline()
+            result = re.search('insertIntoJson<edm4hep::(.+?)Collection>',
+                               cxxline)
+            if result:
+                datatypes_found += ['edm4hep::' + result.group(1)]
 
-    datatypes = datamodel['datatypes']
+    datatypes_found = set(datatypes_found)
 
-    for collname in datatypes:
-        include_string = '#include "' + collname.replace('::', '/') + \
-                         'Collection.h"'
-        include_found = False
-        for line in cxxfile_lines:
-            if include_string in line:
-                include_found = True
+    datatypes = set(datamodel['datatypes'])
 
-        if not include_found:
-            print('ERROR: Following collection not included in the cxx file.')
-            print('       ' + collname)
-            sys.exit(2)
+    if not datatypes.issubset(datatypes_found):
+        missing_datatypes = datatypes - datatypes_found
+        print('ERROR: One or more datatypes are not being converted:')
+        for datatype in missing_datatypes:
+            print('       ' + datatype)
+        sys.exit(2)
 
 
 if __name__ == "__main__":
--- a/test/tools/test_all_collections.py
+++ b/test/tools/test_all_collections.py
@@ -1,0 +1,44 @@
+'''
+Tests if all collection header files are included in the cxx file.
+'''
+
+import sys
+import argparse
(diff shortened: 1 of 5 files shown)


tmadlener:
@kjvbrt should we aim to include this in the tag for the new key4hep release?

kjvbrt:
Hi @tmadlener I would be glad If it is included.

tools/include/edm4hep2json.hxx
```diff
@@ -91,28 +105,207 @@ nlohmann::json processEvent(const podio::Frame& frame, std::vector<std::string>&
...
+      auto& hitCollection = frame.get<edm4hep::TrackerHitCollection>(collList[i]);
+      nlohmann::json jsonColl{
+          {collList[i], {{"collection", hitCollection}, {"collID", coll->getID()}, {"collType", coll->getTypeName()}}}};
+      jsonDict.insert(jsonColl.begin(), jsonColl.end());
+    } else if (coll->getTypeName() == "edm4hep::TrackerHitPlaneCollection") {
+      auto& hitPlaneCollection = frame.get<edm4hep::TrackerHitPlaneCollection>(collList[i]);
+      nlohmann::json jsonColl{
+          {collList[i],
+           {{"collection", hitPlaneCollection}, {"collID", coll->getID()}, {"collType", coll->getTypeName()}}}};
+      jsonDict.insert(jsonColl.begin(), jsonColl.end());
```
tmadlener, line 130: Could this be made into a function template that is then simply called here for every case? That would improve readability quite a bit, I think, something like the following should work:

```cpp
template<typename CollT>
void insertIntoJson(nlohmann::json& jsonDict, const podio::CollectionBase* coll, const std::string& name) {
  const auto* typedColl = static_cast<const CollT*>(coll); // safe to cast, since we have queried the type before
  nlohmann::json jsonColl{
          {name,
           {{"collection", *typedColl}, {"collID", coll->getID()}, {"collType", coll->getTypeName()}}}};
  jsonDict.insert(jsonColl.begin(), jsonColl.end());
}
```
and then making the `if`-`else` chain something like
```cpp
if (coll->getTypeName() == "edm4hep::TrackCollection") {
  insertIntoJson<edm4hep::TrackCollection>(jsonDict, coll, collList[i]);
} else if (/*...*/){
  // ...
}
```

tmadlener (commented):
Are there any tests for this yet? That would make the small refactoring I propose below a bit easier to do, but if you don't have time to fix this now, I think we can also merge this as is and open an issue to keep track of that and address that ASAP after the tag, as the functionality doesn't change.

tmadlener:
Also the failing CI workflow seems to have been ignored in the past and is most likely due to a newer version of HepMC in the stack that is used there.

tools/include/edm4hep2json.hxx
kjvbrt: Yeah, this is much easier to read :) Done

kjvbrt:
I will prepare test to check for all the collections in the current yaml file.

test/tools/test_all_collections.py
```diff
@@ -0,0 +1,44 @@
...
+                         'Collection.h"'
+        include_found = False
+        for line in cxxfile_lines:
+            if include_string in line:
+                include_found = True
+
+        if not include_found:
+            print('ERROR: Following collection not included in the cxx file.')
+            print('       ' + collname)
+            sys.exit(2)
```
tmadlener, line 35: With the set of datatypes that are converted from above, it should then be as simple as doing some `datatypes.issubset(converted_types)` to figure out if there are missing conversions. And then use `datatypes - converted_types` to get a list of all the missing ones.

test/tools/test_all_collections.py
```diff
@@ -0,0 +1,44 @@
...
+        datamodel = yaml.safe_load(yamlfile)
+
+    with open(cxxfile_path, mode='r', encoding="utf-8") as cxxfile:
+        cxxfile_lines = cxxfile.readlines()
+
+    datatypes = datamodel['datatypes']
+
+    for collname in datatypes:
+        include_string = '#include "' + collname.replace('::', '/') + \
+                         'Collection.h"'
```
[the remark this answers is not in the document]
kjvbrt: Done

kjvbrt:
I'm not sure what to do with Podio user data collections. For now they are ignored in the test.

tmadlener (approved):
Nice, thanks for the update.

I think for the user data collections we could cook something up with `std::apply` over the `podio::SupportedUserDataTypes` which would make it automatically correct, but I would postpone that to another PR as that might require a bit of refactoring around these and I think the better direction would actually be to add an abstract `to_json` to `podio::CollectionBase` which would get rid of the whole `if`-`else` chain entirely.

PR #234, opened 2023-11-01
Add a header with constants

jmcarcell:
Useful for https://github.com/HEP-FCC/k4SimGeant4/pull/55 and https://github.com/key4hep/k4EDM4hep2LcioConv/pull/32 and other places so that we can set the same value in all places consistently.

More constants can be added if there are suggestions.

BEGINRELEASENOTES
- Add a constant for CellIDEncoding. Usage:

``` cpp
#include "edm4hep/Constants.h"

std::cout << edm4hep::CellIDEncoding << std::endl;
```

ENDRELEASENOTES


include/Constants.h
andresailer, line 1: Move to `include/edm4hep/Constants.h` also in the source tree

jmcarcell: Done while you were writing this comment

andresailer (approved):
Thanks!

include/edm4hep/Constants.h
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef EDM4HEP_CONSTANTS_H
+#define EDM4HEP_CONSTANTS_H
+
+#include <string>
```
tmadlener, line 22:
```suggestion
#include <string_view>
```

include/Constants.h
andresailer: :bird:s of a :feather:

include/edm4hep/Constants.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace edm4hep {
+  inline constexpr std::string_view CellIDEncoding = "CellIDEncoding";
```
tmadlener, line 25:
```suggestion
  inline static constexpr auto CellIDEncoding = "CellIDEncoding";
```
I think making this a `static const char*` makes it easier to consume for things that take a `const std::string&` as  a `string_view` will lead to

```
error: invalid initialization of reference of type 'const std::string&' {aka 'const std::__cxx11::basic_string<char>&'} from expression of type 'const std::string_view' {aka 'const std::basic_string_view<char>'
```
[compiler explorer example](https://godbolt.org/z/sWaKo5vzY)

include/edm4hep/Constants.h
tmadlener: Actually if we make it `const char*` we don't need any include here.

include/edm4hep/Constants.h
tmadlener: This will also still work if functions take a `string_view` as argument.

tmadlener (approved):

include/edm4hep/Constants.h
jmcarcell: I made it `const char*`; `inline` can be dropped since it's static

PR #235, opened 2023-11-05
Use `FILE_SET` to install the headers in `/include` together with the edm4hep library

jmcarcell:
BEGINRELEASENOTES
- Use `FILE_SET` to install the headers in the top folder together with the library. This also adds them in the `BUILD_INTERFACE`, something that a simple `INSTALL` doesn't do.
- Bump the CMake version of the LCG stacks
- Simplify finding ROOT, don't do environment variable manipulation in CMake

ENDRELEASENOTES

`FILE_SET` requires CMake 2.23 so the version of the LCG stacks has to be bumped. I discovered that it also has the nice property that it complains if any of the sources that are attached to the library are not installed.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,8 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        release-platform: LCG_104/x86_64-el9-clang16-opt
+        container: centos7
+        view-path: /cvmfs/sw-nightlies.hsf.org/key4hep
         run: |
           STARTDIR=$(pwd)
           echo "::group::Build podio"
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        container: almalinux9
+        container: ghcr.io/key4hep/key4hep-images/alma9
         view-path: /cvmfs/sw-nightlies.hsf.org/key4hep
         run: |
           STARTDIR=$(pwd)
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        container: ghcr.io/key4hep/key4hep-images/alma9
+        container: centos7
         view-path: /cvmfs/sw-nightlies.hsf.org/key4hep
         run: |
           STARTDIR=$(pwd)
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        release-platform: LCG_103/x86_64-centos7-clang12-opt
+        release-platform: LCG_104/x86_64-centos7-gcc12-opt
         run: |
           STARTDIR=$(pwd)
           echo "::group::Build podio"
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        release-platform: LCG_103/x86_64-centos7-clang15-opt
+        release-platform: LCG_104/x86_64-el9-clang16-opt
         run: |
           STARTDIR=$(pwd)
           echo "::group::Build podio"
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        container: centos7
+        container: almalinux9
         view-path: /cvmfs/sw-nightlies.hsf.org/key4hep
         run: |
           STARTDIR=$(pwd)
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -47,6 +47,7 @@ jobs:
             -DUSE_EXTERNAL_CATCH2=OFF
           ln -s $(pwd)/compile_commands.json ../
           cd ..
+          pip install pre-commit --user
           pre-commit run --show-diff-on-failure \
             --color=always \
             --all-files
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        release-platform: LCG_104/x86_64-centos7-gcc12-opt
+        release-platform: LCG_103/x86_64-centos7-clang15-opt
         run: |
           STARTDIR=$(pwd)
           echo "::group::Build podio"
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,8 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        container: centos7
-        view-path: /cvmfs/sw-nightlies.hsf.org/key4hep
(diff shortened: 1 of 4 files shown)


edm4hep/CMakeLists.txt
```diff
@@ -7,6 +7,12 @@ PODIO_ADD_DATAMODEL_CORE_LIB(edm4hep "${headers}" "${sources}")
 add_library(EDM4HEP::edm4hep ALIAS edm4hep)
 set_target_properties(edm4hep PROPERTIES ALIAS_GLOBAL true)
 
+file(GLOB_RECURSE top_headers ${PROJECT_SOURCE_DIR}/include/*.h)
+target_sources(edm4hep PUBLIC FILE_SET headers TYPE HEADERS BASE_DIRS ${PROJECT_SOURCE_DIR}/include FILES ${top_headers})
```
tmadlener, line 11: Are these the generated headers that we are re-discovering again with globbing, or are these other headers?

jmcarcell: Just the ones in `/include`; instead of installing them in the top CMakeLists.txt it is done here in L38 together with the edm4hep library

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.23)
```
tmadlener, line 1: This requirement fails approx. half of the CI workflows. Do we need 3.23? Or can we go slightly lower as well?

edm4hep/CMakeLists.txt
tmadlener: Should the generated headers also go into the public headers then?

edm4hep/CMakeLists.txt
```diff
@@ -7,6 +7,12 @@ PODIO_ADD_DATAMODEL_CORE_LIB(edm4hep "${headers}" "${sources}")
...
+target_include_directories(edm4hep PUBLIC
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```
m-fila, line 14: Reading from cmake docs:
> Target properties related to include directories are also modified by target_sources(FILE_SET) as follows:
>
>[INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INCLUDE_DIRECTORIES.html#prop_tgt:INCLUDE_DIRECTORIES)
>
>    If the TYPE is HEADERS, and the scope of the file set is PRIVATE or PUBLIC, all of the BASE_DIRS of the file set are wrapped in [$<BUILD_INTERFACE>](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:BUILD_INTERFACE) and appended to this property.
>
>[INTERFACE_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_INCLUDE_DIRECTORIES)
>
>    If the TYPE is HEADERS, and the scope of the file set is INTERFACE or PUBLIC, all of the BASE_DIRS of the file set are wrapped in [$<BUILD_INTERFACE>](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:BUILD_INTERFACE) and appended to this property.

a `FILE_SET` updates the `INCLUDES_DIRECTORY` so manual setting with `target_include_directories` is redundant. These lines could be removed

edm4hep/CMakeLists.txt
m-fila: The `FILE_SET` seems to be cmake newest take on headers management. 
Alternatively the things could be done the same way as in podio (works with older cmake), namely with the `PUBLIC_HEADER` property:
```cmake
set_property(
    TARGET edm4hep
    APPEND PROPERTY
    PUBLIC_HEADER ${top_headers}
)
```

The `PUBLIC_HEADER` has to be installed in the L37 anyway since this is how podio handles headers (generated headers added to the property by `PODIO_ADD_DATAMODEL_CORE_LIB`).

The advantage of `FILE_SET` over `PUBLIC_HEADER` is preserving the hierarchy in the include directories instead of flattening it. In case of this module the hierarchy is flat anyway.

CMakeLists.txt
jmcarcell: 3.23 is when `FILE_SET` was introduced in CMake: https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets
We can just bump the LCG versions, 103 is on 3.23 and 104 is on 3.26 so any of those should work

edm4hep/CMakeLists.txt
jmcarcell: The `INSTALL_INTERFACE` line is not redundant in principle but it happens the public headers are installed to the include dir and that has the same effect as adding here the INSTALL_INTERFACE, so these lines can be removed.

edm4hep/CMakeLists.txt
jmcarcell: > Should the generated headers also go into the public headers then?

Must? Nope. Would it be nice? Maybe and some cmake code can be simplified; I'll have a look but in a different PR since it's not needed and I think the change comes from or also in the podio macros in that case

tmadlener: Would probably be a podio feature request in any case.

edm4hep/CMakeLists.txt
andresailer: >  the public headers are installed to the include dir and that has the same effect as adding here the INSTALL_INTERFACE

I don't understand why the installation to the include dir has the same effect as adding the INSTALL_INTERFACE. Do you refer to PODIO_GENERATE_DATAMODEL (or the other podio macro) already setting the INSTALL_INTERFACE?

jmcarcell: Both `target_include_directories` and `install` populate `INTERFACE_INCLUDE_DIRECTORIES`:
From the [install documentation](https://cmake.org/cmake/help/latest/command/install.html)
```
INCLUDES DESTINATION

    This option specifies a list of directories which will be added to the [INTERFACE_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_INCLUDE_DIRECTORIES) target property of the <targets> when exported by the [install(EXPORT)](https://cmake.org/cmake/help/latest/command/install.html#export) command. If a relative path is specified, it is treated as relative to the [$<INSTALL_PREFIX>](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:INSTALL_PREFIX).
```
So in this case we already have the install command populating `INTERFACE_INCLUDE_DIRECTORIES`, having also `target_include_directories` with the `INSTALL_INTERFACE` means we'll get the same path duplicated in `INTERFACE_INCLUDE_DIRECTORIES`

jmcarcell: Ahh, this is set from podio:
https://github.com/AIDASoft/podio/blob/master/cmake/podioMacros.cmake#L250

.github/workflows/pre-commit.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
-        release-platform: LCG_99/x86_64-centos7-clang10-opt
+        release-platform: LCG_104/x86_64-el9-clang16-opt
```
tmadlener, line 13: I think in other places we use a Key4hep stack here, since we now have LLVM in there as well.

tmadlener (commented):
How important is this for the new Key4hep release?

It seems the only thing that is missing are a few format complaints from pre-commit / clang-format. Which either haven't been caught by previous versions of clang-format, or the file wasn't checked. Anyhow, this should be fairly straight forward to fix.

.github/workflows/pre-commit.yml
jmcarcell: But then it has to be changed because the run-lcg-view only works with LCG stacks

jmcarcell:
> How important is this for the new Key4hep release?
> 
> It seems the only thing that is missing are a few format complaints from pre-commit / clang-format. Which either haven't been caught by previous versions of clang-format, or the file wasn't checked. Anyhow, this should be fairly straight forward to fix.

For the release it doesn't matter. About the complaints, we can ignore them since I get different ones with the new `.clang-format` so I'll make them when I push the file (probably after the release not to create conflicts in the last PRs)

.github/workflows/pre-commit.yml
tmadlener: Would something specific for pre-commit not work for EDM4hep? Because, e.g. podio uses a key4hep stack for the pre-commit workflow: 

https://github.com/AIDASoft/podio/blob/master/.github/workflows/pre-commit.yml

jmcarcell: Ah yes, I thought this was only used for LCG releases

jmcarcell:
Right now the pre-commit CI works fine but there are changes, probably due to a change in the version of `clang-format`. Since the `.clang-format` is going to change anyway, I'm going to merge this as it is and then make the formatting changes in a different commit

Issue #208, opened 2023-07-04
Andy Buckley's comments on EDM4hep MC structures 


gganis:
Received these comments form Andy Buckley as follow up of his presentation on HepMCv3 at 2nd ECFA workshop on Generators.

The colour flow is the obvious defunct quantity, which has no obvious meaning at detector-level anyway -- I am assuming that EDM4HEP is not attempting to replicate the generator-level event graph, as it doesn't seem to have all the required extra for that. Are you aware of anything actually trying to use colour flows at the EDM level? It would only make sense for afterburner generators, and even then relies on specific internal meanings assigned by the original generator: I know of no generator codes, and certainly no post-generation ones, that attempt to implement afterburner corrections for non-colour-singlet states.

The other points that I had in mind were more comments or initial reactions from the MC perspective on:

- spin: similar to the color flow, what is the use-case for this? I believe passing spins for afterburner decays properly requires a spinor basis to be propagated, as well as the vector: this has caused many problems, as existing standards don't propagate enough information and guesses need to be made. Is there detector physics that can use a spin information, even without this extra basis info?

- float charge: I'm accustomed to the integer 3-charge representation in HepPID and MCUtils. Certainly it makes sense for MC-gen level, to handle quark states. But more generally, why store the charge when there is a standard scheme and decoding software to extract the charge from the PDG ID code? (True fractional-charge exotic particles are very niche.)

- double mass vs float-valued 3-momentum vs double-valued positions: is there a rationale for this? I note that float in the Cartesian representation turned out to not be precise enough for LHC forward physics; not sure if it would be sufficient for the same at e.g. a future ep machine. Also worth thinking about whether an eta-phi fundamental representation could make more sense... was this already done?

Zehvogel:
From a quick search it looks like if we want to stay able to run the `MarlinReco/TrueJet` processor [we need the colour flow](https://github.com/iLCSoft/MarlinReco/blob/38e6130d6246280ad4237c582cb997d529d5946d/Analysis/TrueJet/src/TrueJet.cc#L1273-L1274).

Maybe @MikaelBerggren can comment.

gganis:
Ok, it would not apply to data but it would be used in the interpretation of them.
We need to understand how these needs are handled in other cases, e.g. HepMCv3.

tmadlener:
[...]

PR #237, opened 2023-11-06
Make MCParticle momenta use doubles instead of floats

tmadlener:
BEGINRELEASENOTES
- Switch the `momentum` and `momentumAtEndpoint` of the `MCParticle` from `Vector3f` to `Vector3d`. 
- Add a (deprecated from the beginning) constructor from `Vector3f` to `Vector3d` to ease the transition.

ENDRELEASENOTES

As (partially) discussed in #208 

This breaks building (at least the following):
- [x] https://github.com/HEP-FCC/k4SimGeant4/pull/59
- [x] https://github.com/cepc/CEPCSW/pull/256
- [x] https://github.com/HEP-FCC/k4Gen/pull/25

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -44,11 +44,14 @@ components:
         - double y
         - double z
       ExtraCode:
+        includes: "#include <edm4hep/Vector3f.h>"
         declaration: "
         constexpr Vector3d() : x(0),y(0),z(0) {}\n
         constexpr Vector3d(double xx, double yy, double zz) : x(xx),y(yy),z(zz) {}\n
         constexpr Vector3d(const double* v) : x(v[0]),y(v[1]),z(v[2]) {}\n
         constexpr Vector3d(const float* v) : x(v[0]),y(v[1]),z(v[2]) {}\n
+        [[ deprecated("This constructor will be removed again it is mainly here for an easier transition") ]]\n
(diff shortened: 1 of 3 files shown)


Zehvogel:
do we need real doubles here or should we consider something like ROOTs Double32_t which would have the same behaviour as it had in the lcio world i.e. 64 bit in memory 32 bit on disk?

jmcarcell:
But `Double32_t` is just a typedef to double so writing a different number of bytes has to be handled by the IO code

Zehvogel:
yes we would need to put the logic for that somewhere I guess

edm4hep.yaml
```diff
@@ -44,11 +44,14 @@ components:
         - double y
         - double z
       ExtraCode:
+        includes: "#include <edm4hep/Vector3f.h>"
         declaration: "
         constexpr Vector3d() : x(0),y(0),z(0) {}\n
         constexpr Vector3d(double xx, double yy, double zz) : x(xx),y(yy),z(zz) {}\n
         constexpr Vector3d(const double* v) : x(v[0]),y(v[1]),z(v[2]) {}\n
         constexpr Vector3d(const float* v) : x(v[0]),y(v[1]),z(v[2]) {}\n
+        [[ deprecated("This constructor will be removed again it is mainly here for an easier transition") ]]\n
```
tmadlener, line 53:
```suggestion
        [[ deprecated(\"This constructor will be removed again it is mainly here for an easier transition\") ]]\n
```

PR #238, opened 2023-11-13
[WIP] Simplify python util unittests

tmadlener:
BEGINRELEASENOTES
- Make the python unittests work again after the changes of AIDASoft/podio#514

ENDRELEASENOTES

- [ ] Needs AIDASoft/podio#514


tmadlener:
No longer necessary since the changes that would have made this necessary did not land in podio.

PR #240, opened 2023-11-14
Change EventHeader to EventHeaderName

jmcarcell:
BEGINRELEASENOTES
- Change EventHeader to EventHeaderName; we already have an EventHeaderCollection and its elements are called EventHeader (`edm4hep::EventHeader` more precisely)

ENDRELEASENOTES

Building can work because there is `edm4hep::EventHeader` and the constant is not namespaced but it's going to break at some point, at the very least you can't do things like `using edm4hep::EventHeader`.

### DIFF
--- a/include/edm4hep/Constants.h
+++ b/include/edm4hep/Constants.h
@@ -21,7 +21,7 @@
 
 namespace edm4hep {
 static constexpr const char* CellIDEncoding = "CellIDEncoding";
-static constexpr const char* EventHeader = "EventHeader";
+static constexpr const char* EventHeaderName = "EventHeader";
 } // namespace edm4hep
 
 #endif // EDM4HEP_CONSTANTS_H


Zehvogel (approved):
😩

include/edm4hep/Constants.h
```diff
@@ -21,7 +21,7 @@
 
 namespace edm4hep {
 static constexpr const char* CellIDEncoding = "CellIDEncoding";
-static constexpr const char* EventHeader = "EventHeader";
+static constexpr const char* EventHeaderName = "EventHeader";
```
tmadlener, line 24:
```suggestion
static constexpr const char* eventHeader = "EventHeader";
```

As a "less-typing necessary" variant? Or maybe `EvtHeader`?

Zehvogel: I would prefer clarity over shortness

jmcarcell: I'm all for typing less 🤓 but in this case a mistake from `eventHeader` to `EventHeader` can even make the code compile and have some problems somewhere else and `EvtHeader` is also quite similar. I was going to say let Leonhard decide, but he beat me to it so it's two votes against one. I checked the CI failure and I'm quite sure it's not transient, but it's unrelated. I'll have a look and try to fix it

Issue #245, opened 2023-12-12
Add 4D vector object


osbornjd:
It would be useful to have a generic 4 vector type for (for example) storing position+time information of a particle or object, such as a track vertex. This could also be generalized for use as a generic 4 vector class, for example representing a position + time or momentum + energy.

kkauder:
I believe, relatedly, we'll also need a `edm4eic::CovDiag4f` object.

osbornjd:
There aren't other covariance object in the `edm4hep` yaml file (they are all packed arrays), so I'm not sure if the explicit diagonal covariance objects are something we should just keep in edm4eic.

kkauder:
Correct, I was too hasty. Thought this was a edm4eic issue

tmadlener:
I think this is in general a good idea. I still have a few questions / comments for discussion.

Could this functionality be implemented as something that happens in memory? I.e. we would not necessarily need a dedicated component in the datamodel but rather a handwritten c++ class. The proposal in #248 is effectively mainly `ExtraCode` at the moment.

We already have some [utilities](https://github.com/key4hep/EDM4hep/blob/main/utils/include/edm4hep/utils/kinematics.h) for getting e.g. the 4 momenta from anything that has a `momentum` (and either a `mass` or `energy` but that is somewhat optional). This has the advantage that we immediately have something in our hands that supports all the 4-vector operations. These we would otherwise have to implement ourselves or provide some conversion functionality. This should be extensible to more datatypes that do not have a `momentum` directly but rather something that can be converted into one.

osbornjd:
I am open to implementing this in some other way if you think it is best as I don't have deep knowledge of the edm4hep project, but my initial attempt was to basically model what was implemented for `Vector::3f`. The only reason that I think it would be useful to have a generic 4vector object is that you could in principle have something like `(px,py,pz,E)` or (eta,phi,M,E)` or `(x,y,z,t)` so some utility implementation would have to have the ability to be extensible to a generic range of data types that fit all of these possibilities (I think that is effectively what you said also). So this would have to foresee all use cases of 4-vector operations rather than handing the object to the user and having them do whatever they want with it.

tmadlener:
[...]

PR #248, opened 2023-12-14
feat: add 4 dimensional vector object

osbornjd:
BEGINRELEASENOTES
- Added a Vector4f object for use as a general 4vector. Addresses https://github.com/key4hep/EDM4hep/issues/245

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -9,7 +9,7 @@ components:
     
     #Vector4f with floats
     edm4hep::Vector4f:
-      Description: "Generic 4 vector for storing classical 4D coordinates in memory. 4 vector helper functions are in edm4hep::utils"
+      Description: "Generic vector for storing classical 4D coordinates in memory. Four momentum helper functions are in edm4hep::utils"
       Members:
         - float x
         - float y
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -7,7 +7,6 @@ options:
 
 components:
     
-    #Vector4f with floats
     edm4hep::Vector4f:
       Description: "Generic vector for storing classical 4D coordinates in memory. Four momentum helper functions are in edm4hep::utils"
       Members:
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -6,6 +6,22 @@ options:
   includeSubfolder: True
 
 components:
+    
+    #Vector4f with floats
+    edm4eic::Vector4f:
+      Members:
+        - float x
+        - float y
+        - float z
+        - float t
+      ExtraCode:
+        declaration: "
+        constexpr Vector4f() : x(0),y(0),z(0),t(0) {}\n
+        constexpr Vector4f(float xx, float yy, float zz, float tt) : x(xx),y(yy),z(zz),t(tt) {}\n
+        constexpr Vector4f(const float* v) : x(v[0]),y(v[1]),z(v[2]),t(v[3]) {}\n
+        constexpr bool operator==(const Vector4f& v) const { return (x==v.x&&y==v.y&&z==v.z&&t==v.t) ; }\n
+        constexpr float operator[](unsigned i) const { return *( &x + i ) ; }\n
+      "
 
     # Vector3D with floats
     edm4hep::Vector3f:
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -8,7 +8,8 @@ options:
 components:
     
     #Vector4f with floats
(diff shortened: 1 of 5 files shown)


edm4hep.yaml
```diff
@@ -6,6 +6,23 @@ options:
   includeSubfolder: True
 
 components:
+    
+    #Vector4f with floats
+    edm4hep::Vector4f:
+      Description: "Generic 4 vector for storing classical 4D coordinates in memory. 4 vector helper functions are in edm4hep::utils"
```
tmadlener, line 12:
```suggestion
      Description: "Generic vector for storing classical 4D coordinates in memory. Four momentum helper functions are in edm4hep::utils"
```
Maybe we can avoid some of the confusion by calling this a 4D vector and refer to the Lorentz vector as four momentum (and then just try to do that consistently)

edm4hep.yaml
```diff
@@ -6,6 +6,23 @@ options:
   includeSubfolder: True
 
 components:
+    
+    #Vector4f with floats
```
tmadlener, line 10:
```suggestion
```
I think that is quite obvious from the lines below.

tmadlener (commented):
Thanks. I think this looks good with a few minor corrections. 

There are some vector utilities that have been added by EIC here: [`utils/include/edm4hep/utils/vector_utils.h`](https://github.com/key4hep/EDM4hep/blob/main/utils/include/edm4hep/utils/vector_utils.h). I am not sure if you also want to add this one there, or if you want to wait with that until there is a need to do it.

Usually `pre-commit` will fail because the `README.md` has links to the different types in the YAML file and it checks whether they point to the right place. You can easily fix this via the `scripts/updateReadmeLinks.py` script.

osbornjd:
> There are some vector utilities that have been added by EIC here: [utils/include/edm4hep/utils/vector_utils.h](https://github.com/key4hep/EDM4hep/blob/main/utils/include/edm4hep/utils/vector_utils.h)

I'll work on implementing some of these, as they would probably be useful in the long run

edm4hep.yaml
```diff
@@ -6,6 +6,22 @@ options:
   includeSubfolder: True
 
 components:
+
```
tmadlener, line 9:
```suggestion

```

osbornjd:
I added some basic operator functionality for the 4 vectors. I think that should be sufficient for now and if we need more functionality, I can come back and add it in a future PR.

osbornjd:
What is the best way to test the build failure? This builds for me locally in my environment and all tests pass, so I'm not sure how best to debug. I did re-test with the `precommit run --all-files` and these seem to pass now

tmadlener:
I think the build failures are in the c+17 builds. There is a [`vector_utils_legacy.h`](https://github.com/key4hep/EDM4hep/blob/main/utils/include/edm4hep/utils/vector_utils_legacy.h) that get's dropped in in case there are no concepts available yet:

https://github.com/key4hep/EDM4hep/blob/e48126265a1176ab02dfae472633e94869dfb4e5/utils/include/edm4hep/utils/vector_utils.h#L6-L8

So you would have to implement at least the `vector_XYZ` specializations for `Vector4f` there.

osbornjd:
Looks like everything passes except that a test in an fcc build seg faults. What is the best way for me to test this locally as I don't see these tests in the repo?

tmadlener:
That seems to be unrelated to this change and I am not entirely sure whether the `downstream` workflow is already supposed to be working. It has been a bit finicky in the past.

osbornjd:
Okay sounds good, thanks for letting me know. In that case, if it is not too much of an issue would you mind creating a release after merging this PR? It is not urgent with the holidays coming up, so there is absolutely no rush on this

tmadlener (approved):

tmadlener:
I will wait with merging until after the break even though this should not break anything. If nothing has happened after a few days in January, can you ping me again? A new tag is no problem.

osbornjd:
Sounds great, will do. Thanks for your help!

osbornjd:
Unless there are any further changes requested (and now since it is after the holiday break), are there any objections to merge this and create a new release?

tmadlener (approved):

tmadlener:
There is a new tag and there is also a PR in spack to pick it up: https://github.com/spack/spack/pull/41987

osbornjd:
Thanks very much for the new tag!

wdconinc:
Merged in spack.

PR #249, opened 2024-01-08
Fix formatting in components table

m-fila:
BEGINRELEASENOTES
- Fixed formatting of components table in readme

ENDRELEASENOTES

After adding the `Vector4f` in  #248  the formatting of 'components' table in readme was broken. The table is supposed to have three columns, while in some row four cells were given causing rendering problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,10 @@ A generic event data model for future HEP collider experiments.
 
 | | | |
 |-|-|-|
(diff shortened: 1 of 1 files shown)


tmadlener (approved):
Thanks for fixing this. I didn't think about checking this when merging the PR.

Not sure if it is worth to try and implement a check for this as well that can be run in CI. Let's see how often we get bitten by this ;)

PR #252, opened 2024-01-19
Use interface types for TrackerHit

### LINKED ISSUE #218: Handling of different `TrackerHit` types
Currently there are two different types of `TrackerHit` types: `TrackerHit` and `TrackerHitPlane`. (LCIO knows [even more](https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHit.html)). However, the `Track` only has a relation to `TrackerHit`, so it is not possible to relate `TrackerHitPlane`s to `Track`s at the moment.

This will need support from the podio side. One proposal is https://github.com/AIDASoft/podio/pull/215

tmadlener:
BEGINRELEASENOTES
- Make the `edm4hep::TrackerHit` an `interface` type and use that in the `edm4hep::Track`.
- Rename the current `edm4hep::TrackerHit` to `edm4hep::TrackerHit3D`
- **This is a breaking change. However, we currently do not plan to implement schema evolution for this**

ENDRELEASENOTES

- [x] Needs AIDASoft/podio#516
Breaking changes downstream:
- [x] https://github.com/key4hep/k4EDM4hep2LcioConv/pull/45
- [x] https://github.com/key4hep/k4FWCore/pull/174
- [x] https://github.com/AIDASoft/DD4hep/pull/1219 (DDDigi also uses the TrackerHit and needs to be updated accordingly.)
- [x] https://github.com/key4hep/k4MarlinWrapper/pull/168
- [x] https://github.com/key4hep/k4SimDelphes/pull/113
- [x] https://github.com/HEP-FCC/k4SimGeant4/pull/62
- [x] https://github.com/HEP-FCC/FCCAnalyses/pull/345
- [x] CEPCSW breaks (https://github.com/cepc/CEPCSW/issues/263)
- [x] Update documentation (add an `interfaces` section?)

This is one of a few rather breaking changes that we plan to do in the near future towards an EDM4hep version 1.0.

Fixes #218

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -396,7 +396,7 @@ datatypes:
       - edm4hep::TrackState trackStates  // track states
       - edm4hep::Quantity dxQuantities // different measurements of dx quantities
     OneToManyRelations:
-      - edm4hep::TrackerHit3D trackerHits  // hits that have been used to create this track
+      - edm4hep::TrackerHit trackerHits  // hits that have been used to create this track
       - edm4hep::Track tracks            // tracks (segments) that have been combined to create this track
 
  #---------- Vertex
@@ -475,7 +475,7 @@ datatypes:
     Members:
       - float weight              // weight of this association
     OneToOneRelations:
-     - edm4hep::TrackerHit3D rec    // reference to the reconstructed hit
+     - edm4hep::TrackerHit rec    // reference to the reconstructed hit
      - edm4hep::SimTrackerHit sim // reference to the simulated hit
 
   edm4hep::MCRecoTrackerHitPlaneAssociation:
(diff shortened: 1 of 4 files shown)


edm4hep.yaml
```diff
@@ -591,3 +591,20 @@ datatypes:
       - edm4hep::HitLevelData hitData   //hit level data
     OneToOneRelations:
       - edm4hep::Track track           //the corresponding track.
+
+interfaces:
+  edm4hep::TrackerHit:
+    Description: "Tracker hit interface class"
+    Author: "Thomas Madlener, CHEP"
```
tmadlener, line 598:
```suggestion
    Author: "Thomas Madlener, DESY"
```

edm4hep.yaml
```diff
@@ -594,3 +594,20 @@ datatypes:
...
+    Author: "Thomas Madlener, DESY"
+    Members:
+      - uint64_t cellID // ID of the sensor that created this hit
+      - int32_t type // type of the raw data hit
+      - int32_t quality // quality bit flag of the hit
+      - float time [ns] // time of the hit
+      - float eDep [GeV] // energy deposited on the hit
+      - float eDepError [GeV] // error measured on eDep
+      - edm4hep::Vector3d position [mm] // hit position
+      - std::array<float, 6> covMatrix // covariance of the position (x,y,z) stored as lower triangle matrix, i.e. cov(x,x), cov(y, x), cov(y, y), cov(z, x), cov(z, y), cov(z, z)
```
tmadlener, line 610:
```suggestion
```

Remove the 3D covariance matrix as discussed during the last EDM4hep meeting (Jan 23).

tmadlener: Done

jmcarcell:
I'm not sure how to fix the CI for this one. So it fails because it picks up the FCCAnalyses dictionaries and headers and there something wants `TrackerHitData.h` which doesn't exist anymore with this PR. Why ROOT does this when it's not needed I don't know. Locally the tests pass for me without FCCAnalyses so one option would be to remove FCCAnalyses in the environment for the tests as they should be only related to EDM4hep. However this is useful information since people using EDM4hep in Key4hep will have FCCAnalyses in the environment, so we can be aware if something is failing. I think I'll just modify the environment and add a test to the tests that run post-build so that we would be aware if there were issues like this one.

tmadlener:
The problem here is that `LD_LIBRARY_PATH` is still in the test environment. `FCCAnalyses` from the nightlies is still built against the older EDM4hep, so it still has a `TrackerHitData`. ROOT picks up things pretty indiscriminantely from there. This would be solvable in EDM4hep probably by constructing the test environment to contain exactly what it needs, because there is no FCCAnalyses in the LCG stacks. LCG stack environments are pretty impossible to deal with in these cases because there is almost always something that pulls in `<lcg-view-path>/lib[64]`.

PR #254, opened 2024-01-22
Add weights vector to EventHeader and constant for storing them

### LINKED ISSUE #242: Ways to store multiple event weights
Some event generators produce more than one event weight. However, currently the `EventHeader` only has one `weight` so storing multiple event weights is not easily possible. I had a brief discussion with @selvaggi about different ways to this and some considerations to be taken into account. The possibilities are

1. Store the weights as `UserDataCollection`. This would require no real changes to EDM4hep or podio, but would require the introduction of a new convention for naming this collection.
2. Add a `VectorMember` of `double` to the `EventHeader`. This would potentially require some schema evolution developments for podio (needs confirmation) but would not require any new naming convention.
3. Store the weights as key - value pairs in the Frame parameters. This would also not require any changes to EDM4hep or podio, but would make the weight values rather hard to obtain with plain root / RDataFrame means.

The first two options would also require to store the names of the weights separately. However, since those are the same for all events, they could just be put into the file level metadata for which all the necessary pieces are already in place. The one thing that would then be necessary in, e.g. FCCAnalyses would be to construct a mapping of the names to the indices in either the `UserDataCollection` or the `VectorMembers` for the corresponding weight. That should be fairly straight forward we think as it only needs to be done once, and can be entirely hidden from the user, unless @kjvbrt does not agree there.

Our favorite currently would be to go with option 2, i.e. an additional `VectorMember` in the `EventHeader` as it seems to be the most stable and natural solution.

As a side note: LCIO achieves this with its version of option 3.

Potentially related: #241

tmadlener:
BEGINRELEASENOTES
- Add a `double weight` `VectorMember` to the `EventHeader` to allow storing multiple weights.
  - **This should not break any existing code as we simply add a new member. However, this will break the possibility of reading existing files through podio**.
- Introduce the `edm4hep::EventWeights` constant to have a consistent name for storing / accessing them in file level metadata.

ENDRELEASENOTES

- Fixes #242 
- [x] Make `weight` a `double` as well? -> Yes

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -149,7 +149,7 @@ datatypes:
       - int32_t eventNumber    //event number
       - int32_t runNumber      //run number
       - uint64_t timeStamp     //time stamp
-      - float weight       // event weight
+      - double weight       // event weight
     VectorMembers:
       - double weights       // event weights in case there are multiple. **NOTE that weights[0] might not be the same as weight!** Event weight names should be stored using the edm4hep::EventWeights name in the file level metadata.
 
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -150,6 +150,8 @@ datatypes:
       - int32_t runNumber      //run number
       - uint64_t timeStamp     //time stamp
       - float weight       // event weight
+    VectorMembers:
+      - double weights       // event weights in case there are multiple. **NOTE that weights[0] might not be the same as weight!** Event weight names should be stored using the edm4hep::EventWeights name in the file level metadata.
 
 
   edm4hep::MCParticle:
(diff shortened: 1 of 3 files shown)


tmadlener:
As discussed during the [meeting on Jan 23](https://indico.cern.ch/event/1371141/) we also make the nominal weight of the `EventHeader` a `double` for consistency.

edm4hep.yaml
```diff
@@ -150,6 +150,8 @@ datatypes:
       - int32_t runNumber      //run number
       - uint64_t timeStamp     //time stamp
       - float weight       // event weight
```
tmadlener, line 152:
```suggestion
      - double weight       // event weight
```

PR #257, opened 2024-01-25
Add Python formatting CMake targets (black, flake8)

graeme-a-stewart:
Add CMake targets for running black and flake8 on Python source files
CMake will look to find the relevant binaries and setup the targets if they are there, if they are not found a STATUS message is printed
Targets are excluded from "all" to avoid running them by accident

BEGINRELEASENOTES
Add CMake targets for running black and flake8 on Python source files
ENDRELEASENOTES

This is basically the same PR as https://github.com/AIDASoft/podio/pull/544

### DIFF
--- a/cmake/pythonFormat.cmake
+++ b/cmake/pythonFormat.cmake
@@ -4,7 +4,8 @@
 
 
 # Get all our Python files
-file(GLOB_RECURSE ALL_PYTHON_FILES ${PROJECT_SOURCE_DIR}/*.py)
+file(GLOB_RECURSE ALL_PYTHON_FILES ${PROJECT_SOURCE_DIR}/python/*.py 
+  ${PROJECT_SOURCE_DIR}/scripts/*.py ${PROJECT_SOURCE_DIR}/test/*.py)
 
 # Black is rather simple because there are no options...
 find_program(BLACK_EXECUTABLE black)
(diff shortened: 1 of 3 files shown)


cmake/pythonFormat.cmake
```diff
@@ -0,0 +1,34 @@
...
+else()
+    message(STATUS "Failed to find black executable - no target to run black can be set")
+endif()
+
+find_program(FLAKE8_EXECUTABLE flake8)
+if(FLAKE8_EXECUTABLE)
+    add_custom_target(
+        flake8
+        COMMAND ${FLAKE8_EXECUTABLE}
+        --config=${CMAKE_CURRENT_SOURCE_DIR}/.flake8
```
tmadlener, line 27:
```suggestion
        --config=${PROJECT_SOURCE_DIR}/.flake8
```

However, this file does not yet exist in EDM4hep

cmake/pythonFormat.cmake
```diff
@@ -0,0 +1,34 @@
+# Additional target to run python linters and formatters on python scripts
+#
+# Requires black/flake8 to be available in the environment
+
+
+# Get all our Python files
+file(GLOB_RECURSE ALL_PYTHON_FILES ${PROJECT_SOURCE_DIR}/*.py)
```
tmadlener, line 7:
```suggestion
file(GLOB_RECURSE ALL_PYTHON_FILES ${PROJECT_SOURCE_DIR}/*.py)
```
This will also pick up a potential `install` folder from a (previous) build.

cmake/pythonFormat.cmake
graeme-a-stewart: Done!

cmake/pythonFormat.cmake
graeme-a-stewart: Yes, now restricted to our own Python source directories, so it's a but cleaner

tmadlener:
I just realized that we have neither a flake8 nor a black configuration yet for EDM4hep. If it's OK with you, I would simply add them with this PR (and also apply the necessary style fixes).

PR #260, opened 2024-02-01
Remove the edm4hep::ObjectID from the components

### LINKED ISSUE #259: Remove the `edm4hep::ObjectID` from the datamodel
Currently edm4hep defines its own `ObjectID`:
https://github.com/key4hep/EDM4hep/blob/66c90fc6d45fdb26793510400ff180d070065eff/edm4hep.yaml#L110-L120

This is slightly different from the podio one which basically boils down to
```cpp
struct ObjectID {
  int index;
  uint32_t collectionID;
};
```

As far as I can tell this is actually unused and we should be able to achieve the same (intended) functionality.

tmadlener:
BEGINRELEASENOTES
- Remove the `edm4hep::ObjectID` component. Fixes #259 
- Remove its usage from the `edm4hep::TrackerHit` and `edm4hep::TrackerHitPlane`
- **This is a breaking change both in code and existing files**. Existing usages might switch to use the `podio::ObjectID` to have effectively the same functionality.

ENDRELEASENOTES

- [x] Breaks in CEPCSW
- [x] Fix README.md links once this becomes ready to be merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,21 +21,21 @@ A generic event data model for future HEP collider experiments.
 
 | | | |
 |-|-|-|
-| [EventHeader](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L136)         | [MCParticle](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L147)        | [SimTrackerHit](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L215)         |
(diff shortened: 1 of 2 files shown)


tmadlener:
The `rawHits` are used extensively in CEPCSW!

jmcarcell:
I think this is ready to go?

jmcarcell:
I'll enable auto-merge. It will take a while though :computer: :fire:

tmadlener:
Wait, let me quickly check if this does indeed not break anything else than CEPCSW. Will re-enable automerge once that is confirmed.

PR #256, opened 2024-01-23
Rename MCParticle relations to particle to avoid clashes with type names

tmadlener:
BEGINRELEASENOTES
- Rename `MCParticle` relations to `particle` in `SimTrackerHit` and `SimPrimaryIonizationCluster` to avoid clashes with the `MCParticle` typename.
  - Add `getMCParticle` and `setMCParticle` via `ExtraCode` to keep everything working.
  - **This will break the reading of existing files via podio utilities**

ENDRELEASENOTES

- [x] Check if everything still builds with the ExtraCode

Fixes #247

PR #261, opened 2024-02-01
V1.0 breaking changes

tmadlener:
**This is a PR that is not intended to be merged!** It collects all the upcoming breaking changes in one place for easier testing of the final result after all of them have been merged. It currently contains the changes from:

- #260 
- #256 
- #254
- #252 

**More changes are still to come and will show up here as soon as they are ready**. Please use the dedicated PRs for mentioning specific issues that are caused / raised by them.

@wdconinc this is a sort of WIP of the branch we discussed at the last Key4hep meeting. 

Also tagging @BrieucF @kjvbrt and @mirguest as these will almost certainly also affect you.


jmcarcell:
Should we have a branch like this one in every affected repo? Otherwise it's impossible to test it, and should have the same name obviously

jmcarcell:
At this point this PR is not useful since all but one of the PRs have been merged and this one isn't up to date

wdconinc:
This is what causes us to move more datatypes into EDM4eic instead of using EDM4hep :shrug:

PR #263, opened 2024-02-06
Change ROOTFrame{Writer,Reader} to ROOT{Writer,Reader}

jmcarcell:
BEGINRELEASENOTES
- Change ROOTFrame{Writer,Reader} to ROOT{Writer,Reader} following https://github.com/AIDASoft/podio/pull/549

ENDRELEASENOTES

When the #ifs are removed we should specify a requirement of podio 0.99 or greater


tmadlener:
> When the #ifs are removed we should specify a requirement of podio 0.99 or greater

Yes, one of (if not the next tag of EDM4hep) will depend on either podio >= 0.99 or even podio >= 1.0 I would say.

PR #264, opened 2024-02-06
Delete build workflow since we have another one for key4hep

jmcarcell:
BEGINRELEASENOTES
- Delete build workflow since we have another one for key4hep that covers builds for nightlies, releases and all the operating systems we support

ENDRELEASENOTES


tmadlener:
Will this automatically build with c++20 for the nightlies?

jmcarcell:
Yes: [downstream-build](https://github.com/key4hep/key4hep-actions/tree/main/downstream-build)

tmadlener (approved):

tmadlener:
We actually lost some testing with this. Previously we also built the tests that check downstream usage, which we now no longer do.

We still build them with the LCG based CI workflows:
https://github.com/key4hep/EDM4hep/blob/f3efa0ad302e56b8aadade47077913e7bfe004ee/.github/workflows/lcg_linux_with_podio.yml#L62-L71

PR #266, opened 2024-02-07
Revert "Make MCParticle momenta use doubles instead of floats"

tmadlener:
Reverts key4hep/EDM4hep#237

BEGINRELEASENOTES

- Revert making MCParticle momenta double precision, since this breaks without schema evolution. (Reverts key4hep/EDM4hep#237)
  - All particle momenta are effectively zeroed out at the moment.

ENDRELEASENOTES

@wdconinc in case you have picked up `edm4hep@0.10.4` already via spack you will most likely run into issues with that.

**We will bring these changes back in the breaking set of things for v1.0.**

Fixes https://github.com/key4hep/k4MarlinWrapper/issues/170


wdconinc:
Thanks. We're still on 0.10.3.

tmadlener:
OK. I almost figured you might not have caught up yet, because you would have noticed a bit earlier probably ;) I will make a 0.10.5 tag after this is merged.

PR #267, opened 2024-02-08
Make MCParticle momenta use doubles instead of floats

tmadlener:
BEGINRELEASENOTES
- Bring back MCParticle momenta as doubles. See #237 and #266

ENDRELEASENOTES


jmcarcell:
This was auto-merged but not all the CI passed, there seems to be something broken in the wrapper that I think we already knew about:
https://github.com/key4hep/EDM4hep/actions/runs/8001763830/job/21853603427

tmadlener:
I think this is based on alma9 and things should work there. The break in the wrapper is only present for centos7 on the nightlies.

jmcarcell:
That CI is running on Alma 9, so I think there is something wrong that doesn't always happen

tmadlener:
This might actually just be because the file that is read in some tests can no longer be read due to all the changes. We will have to check that and create a new one once we are done.

PR #268, opened 2024-02-08
Reverse the direction of the ParticleID relation(s)

### LINKED ISSUE #217: Direction of inter-object relations
The immutability of objects that are read from file (or that are the outputs of an algorithm) is one of the most central aspects of podio generated EDMs. This requirement is/was no present in LCIO on which many aspects of the current definition build. In many of the existing workflows from LCIO it is assumed that it is possible to still add relations **after** the original objects have  been created. For example,

- Adding `ParticleID` objects to existing `ReconstructedParticle`s (and `Cluster`s)
- Adding a `startVertex` to an existing `ReconstructedParticle`
- Updating dE/dx values after a track refit
- Refitting a track with an updated PID (and a different) mass and updating the track information after this fit

For some of these use cases it would potentially make sense to *reverse* the direction of the intrinsic one-to-one or one-to-many relations, e.g. Add a `ReconstructedParticle` to the `ParticleID` and remove the `ParticleID`s from the `ReconstructedParticle`. In this way we would effectively always require that a "finished" reco particle is present to run any particle ID on top of that, and then the PID result would know for which reco particle it was run. 

Similar arguments can be made for other relations, and we would probably need to decide on a case-by-case basis on what to do best.

This issue should serve as a board for summarizing discussions and potentially fleshing out some of the details.

### LINKED ISSUE #251: Remove the `particleIDUsed` from `ReconstructedParticle`
Unused in LCIO

tmadlener:
BEGINRELEASENOTES
- Remove the relations from `Cluster` to `ParticleID`
- Remove the relations from `ReconstructedParticle` to `ParticleID`
- Add a `OneToOneRelation` from `ParticleID` to a `ReconstructedParticle`
- **This is a breaking change for both existing files and interfaces. There will be no schema evolution for this change**
- Add a new `edm4hep::utils::PIDHandler` and some related utility functionality to help with handling the necessary meta data.

ENDRELEASENOTES

Fixes #251, Fixes #217 

Associated PRs (need to be merged at the same time to not break the nightlies completely)
- [key4hep/k4EDM4hep2LcioConv#51](https://github.com/key4hep/k4EDM4hep2LcioConv/pull/51)
- [key4hep/k4MarlinWrapper#181](https://github.com/key4hep/k4MarlinWrapper/pull/181)
- [key4hep/k4SimDelphes#117](https://github.com/key4hep/k4SimDelphes/pull/117)

Open questions (so far):
- [x] Does the `ParticleID` still need the `type` and `algorithmType` members. In principle at least the latter now can be entirely encoded in the ParticleID collection, since in principle each algorithm can (should?) produce its own collection?
  - Probably not strictly, but we keep it around for now to not complicate this further.


tmadlener:
I have added a first version of a `PIDHandler` class that can reverse the direction of the ParticleID -> ReconstructedParticle relation.

Since `EDM4HEP::utils` now becomes an actual shared library, I have renamed the target because otherwise we will just get a `libutils.so` which might clash. To ease the transition for now, I add back the `EDM4HEP::utils` target for downstream consumers, but we could also consider simply breaking things entirely.

utils/include/edm4hep/utils/ParticleIDUtils.h
```diff
@@ -0,0 +1,48 @@
...
+    handler.addColl(coll);
+    (handler.addColl(pidColls), ...);
+    return handler;
+  }
+
+  /// Add the information from one ParticleIDCollection to the handler
+  void addColl(const edm4hep::ParticleIDCollection& coll);
+
+  /// Retrieve all ParticleIDs that are related to the passed
+  /// RecontstructedParticle
```
jmcarcell, line 43:
```suggestion
  /// ReconstructedParticle
```

utils/include/edm4hep/utils/ParticleIDUtils.h
```diff
@@ -0,0 +1,96 @@
...
+struct ParticleIDMeta {
+  std::string algoName{};                ///< The name of the algorithm
+  int32_t algoType{0};                   ///< The (user defined) algorithm type
+  std::vector<std::string> paramNames{}; ///< The names of the parameters
+};
+
+/// Get the index of the parameter in the passed ParticleID meta info
+std::optional<int> getParamIndex(const ParticleIDMeta& pidMetaInfo, const std::string& param);
+
+/// Utility class to invert the ParticleID to ReconstructedParticle relation
```
andresailer, line 27: Could you add a code example how the PIDHandler is supposed to be used?
Created, filled, information extracted?

tmadlener: Done. Visible [here](https://github.com/tmadlener/EDM4hep/blob/particleid-reversal/doc/PIDHandler.md)

andresailer (commented):
Thanks for the documentation! It only allows me to offer a minor nitpick.

doc/PIDHandler.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+### Setting meta information for a collection
+
+If you have a `ParticleIDCollection` and want to persist some meta information
+for it the following snippet shows you how to do that
+
+```cpp
+// Create ParticleID meta information for an algorithm
+const auto tofMeta = {"TimeOfFlight", 42, {"time", "time_error"}};
```
tmadlener, line 195:
```suggestion
const auto tofMeta = edm4hep::utils::ParticleIDMeta{"TimeOfFlight", 42, {"time", "time_error"}};
```

doc/PIDHandler.md
```diff
@@ -0,0 +1,229 @@
...
+  `event` Frame and the one that is passed to `PIDHandler::setAlgoInfo`!
+
+### Setting meta information using a collection name
+
+If you want to set the meta information for a `ParticleIDCollection` that you no
+longer have mutable access to, this can be done via
+
+```cpp
+// Create ParticleID meta information for an algorithm
+const auto tofMeta = {"TimeOfFlight", 42, {"time", "time_error"}};
```
tmadlener, line 219:
```suggestion
const auto tofMeta = edm4hep::utils::ParticleIDMeta{"TimeOfFlight", 42, {"time", "time_error"}};
```

utils/src/ParticleIDUtils.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  return pids;
+}
+
+std::optional<edm4hep::ParticleID> PIDHandler::getPID(const edm4hep::ReconstructedParticle& reco,
+                                                      int32_t algoType) const {
+  for (const auto& pid : getPIDs(reco)) {
+    if (pid.getAlgorithmType() == algoType) {
+      return pid;
```
jmcarcell, line 82:
```suggestion
  const auto& [begin, end] = m_recoPidMap.equal_range(reco);
  for (auto it = begin; it != end; ++it) {
    if (it->getAlgorithmType() == algoType) {
      return *it;
```

The vector of PIDs doesn't need to be created

doc/PIDHandler.md
```diff
@@ -0,0 +1,229 @@
...
+Alternatively it can be created as an empty Frame and then just written using
+the `podio::Category::Metadata` category name.
+
+**Finally most of the examples assume that the desired values were found and
+simply get the `value` from the returned `std::optional` directly.** This means
+that most of the times you will see lines like this
+
+```cpp
+const auto value = pidHandler.getAlgoType("someAlgo").value();
+```
```
jmcarcell, line 88: The problem here is that the exception messages are very cryptic:
```
terminate called after throwing an instance of 'std::bad_optional_access'
  what():  bad optional access
zsh: IOT instruction (core dumped)  ./a.out
```
so encouraging not checking is not great. At the same time I think this part is too verbose, we don't need to teach `std::optional` here.

doc/PIDHandler.md
```diff
@@ -0,0 +1,229 @@
+# PIDHandler introduction and usage
```
jmcarcell, line 1: Maybe somewhere it should say that it's not mandatory to use the PIDHandler and one can iterate manually? For example if each reco particle needs to be accessed once for a given PID algorithm, then it's not necessary to build a map.

tmadlener: Good point. Done.

doc/PIDHandler.md
tmadlener: Maybe I have to reword this a bit. The section here introduces the assumptions we make in the examples below. This should not be read as encouraging using `value` directly. If we do the proper checks, the examples get rather long.

utils/src/ParticleIDUtils.cc
tmadlener: I was going for some code reuse, but good point. Changed.

Issue #270, opened 2024-02-20
The conversion operator is not being used in python to go from Mutable to the non-mutable type


jmcarcell:
For example, the following code will fail:
``` python
import podio
import edm4hep
particles = edm4hep.MCParticleCollection()
parent_particle = particles.create()
daughter_particle = particles.create()
parent_particle.addToParents(daughter_particle)
```

with the error:
```
TypeError: void edm4hep::MutableMCParticle::addToParents(edm4hep::MCParticle) =>
    TypeError: could not convert argument 1
```

However if one does interactively `daughter_particle.operator<TAB>` it will get expanded to 
```daughter_particle.operator MCParticle```
 but with a space in between it's not valid syntax. We can try: 
```daughter_particle.__getattribute__('operator MCParticle')()```
 but this doesn't work either.

There is a workaround for this:

``` python
import cppyy.ll
...
    parent_particle.addToParents(cppyy.ll.cast['edm4hep::MCParticle'](daughter_particle))
```
(`static_cast` can also be used, `cast` is simply a C-like cast)

More information about operators in cppyy: https://cppyy.readthedocs.io/en/latest/classes.html#operators and https://cppyy.readthedocs.io/en/latest/type_conversions.html#operators

It looks to me this is something on the cppyy side, that the cppyy wrapped objects won't convert automatically since we have seen that when we cast ourselves the operator works fine.

tmadlener:
Yes, that is definitely a good idea regardless. Another issue might be that the cppyy that comes with ROOT could not be the latest version of cppyy.

jmcarcell:
Ugh the version from ROOT is almost 4 years old...: https://github.com/wlav/cppyy/releases/tag/cppyy-1.6.2

m-fila:
Conversion function doesn't work with the newest standalone cppyy either:

```python
>>> import cppyy
>>> import cppyy.ll
>>> cppyy.__version__
'3.1.2'
>>> cppyy.cppdef("""
... 
... namespace ns {
... struct To{};
... 
... struct From{
...     operator To() const {return To();} 
... };
... 
... int foo(const To&) {return 0;}
... 
... }""")
True
>>> f = cppyy.gbl.ns.From()
>>> foo = cppyy.gbl.ns.foo
>>> foo(f)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: int ns::foo(const ns::To&) =>
    TypeError: could not convert argument 1

```

[...]

PR #272, opened 2024-02-21
Add a script to create an EDM4hep file making use of the complete datamodel

jmcarcell:
BEGINRELEASENOTES
- Add a script to create an EDM4hep file. The script is in python and can be called like `python createEDM4hepFile.py`. I have tried to make use of every member and relation in the data model. The purpose is to have something that can create quickly an EDM4hep file and because the script is in python no compilation is needed, hopefully lowering the barrier to get to an EDM4hep file. People can adapt it to their needs or use it as an example for testing, debugging, etc.

ENDRELEASENOTES
For the question, "should this be a test?" I don't have a strong opinion but I think it would have less visibility than in `/scripts`


peremato:
Great! well done 

> For the question, "should this be a test?" I don't have a strong opinion but I think it would have less visibility than in /scripts
A test can be created that uses the script in /scripts

Two  questions (perhaps are trivial)
- how do you choose the ROOT format (TTree or RNTuple)?
- Why do you need `cppyy.ll.static_cast`?

jmcarcell:
> Great! well done
> 
> > For the question, "should this be a test?" I don't have a strong opinion but I think it would have less visibility than in /scripts
> > A test can be created that uses the script in /scripts
> 
> Two questions (perhaps are trivial)
> 
>     * how do you choose the ROOT format (TTree or RNTuple)?
> 
>     * Why do you need `cppyy.ll.static_cast`?

I added an option `--rntuple` to save to RNTuples. For the second question it seems it's a problem in `cppyy` maybe, see: https://github.com/key4hep/EDM4hep/issues/270 (the conversion operator is not being used from the mutable to the non-mutable types)

tmadlener (commented):
We should add a run of this script to CI, to make sure it always works.

.github/workflows/generate-edm4hep-file.yaml
```diff
@@ -0,0 +1,45 @@
+name: Generate an EDM4Hep file
```
tmadlener, line 1: Can we not attach this to the usual tests we run? Resp. what am I missing that speaks against this? File size, run time,...?

jmcarcell: Changed now to a test

test/CMakeLists.txt
```diff
@@ -6,12 +6,16 @@ function(set_test_env _testname)
   set_property(TEST ${_testname} APPEND PROPERTY ENVIRONMENT
       LD_LIBRARY_PATH=$<TARGET_FILE_DIR:edm4hep>:$<TARGET_FILE_DIR:podio::podio>:$<$<TARGET_EXISTS:edm4hepRDF>:$<TARGET_FILE_DIR:edm4hepRDF>:>$<TARGET_FILE_DIR:ROOT::Core>:$ENV{LD_LIBRARY_PATH}
       ROOT_INCLUDE_PATH=${PROJECT_SOURCE_DIR}/edm4hep:${PROJECT_SOURCE_DIR}/utils/include:$ENV{ROOT_INCLUDE_PATH}
+      PYTHONPATH=${PROJECT_SOURCE_DIR}/python:$ENV{PYTHONPATH}
```
tmadlener, line 9: We either need to add the podio that we have used for building here as well to `PYTHONPATH`, or we have to do it in the workflow file in the step where we build and install podio. CI is partially failing, because it picks up an older version of podio from the LCG stack:

https://github.com/key4hep/EDM4hep/actions/runs/10045234420/job/27762066625?pr=272#step:4:817

Issue #262, opened 2024-02-06
Edm4hep files contaning vertex collections incompatible with root RDataFrame.Describe()


Zehvogel:
root's RDataFrame `.Describe()` feature fails with edm4hep files containing vertex collections with the following error message:

```
cppyy.gbl.std.runtime_error: ROOT::RDF::RDFDescription ROOT::RDF::RInterfaceBase::Describe() =>
    runtime_error: TTree leaf BuildUpVertices.covMatrix[6] has both a leaf count and a static length. This is not supported.
```

It is probably not our fault but I wanted to submit the issue here first

 - OS version: alma9
 - Package version: `source /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-02-05/x86_64-almalinux9-gcc11.3.1-opt/key4hep-stack/2024-02-05-osrelo/setup.sh`
 - Reproduced by: 
downloading https://cernbox.cern.ch/s/71P17DwMNplpruv
```python
import ROOT
df = ROOT.RDataFrame("events", "cld_enuW_1evt_REC_edm4hep.root")
df.Describe()
```

tmadlener:
This looks like a root issue, where the minimal reproducer should be to effectively write a branch with an `std::array`. It could also be that it fails only if the `std::array` is part of a `struct` as that is what it is in our case.

Zehvogel:
Looks like a reproducer is not so easy to produce...

I made a struct
```c++
struct test {
    std::array<float, 6> arr;
    ROOT::VecOps::RVec<float> vec;
    ROOT::VecOps::RVec<ROOT::VecOps::RVec<float>> vecvec;
    std::array<std::array<float, 6>,1> arrarr;
    ROOT::VecOps::RVec<std::array<float,6>> vecarr;
}
```
and use it like this:

```python
import ROOT

df = ROOT.RDataFrame(10)

code = """
struct test {
    std::array<float, 6> arr;
    ROOT::VecOps::RVec<float> vec;
    ROOT::VecOps::RVec<ROOT::VecOps::RVec<float>> vecvec;
    std::array<std::array<float, 6>,1> arrarr;
    //ROOT::VecOps::RVec<std::array<float,6>> vecarr;
};
"""
ROOT.gInterpreter.Declare(code)

df = df.Define("test", "test()")
print(df.Describe())
df.Snapshot("test_tree", "test_file.root")
df2 = ROOT.RDataFrame("test_tree", "test_file.root")
print(df2.Describe())
[...]

PR #273, opened 2024-02-22
Rename type to PDG in ReconstructedParticle

jmcarcell:
BEGINRELEASENOTES
- Rename type to PDG in ReconstructedParticle

ENDRELEASENOTES

Fixes https://github.com/key4hep/EDM4hep/issues/262


tmadlener:
This might break a few things downstream. Checking which at the moment.

Issue #180, opened 2022-10-18
Inconsistent naming of energy deposit in different types


tmadlener:
Except we have both "EDep" and "eDep" in existing types already...

_Originally posted by @andresailer in https://github.com/key4hep/EDM4hep/pull/179#pullrequestreview-1145216205_

andresailer:
https://github.com/key4hep/EDM4hep/blob/f0d153f698c1ca3ad3c5679931b684eb296ad25e/edm4hep.yaml#L197
https://github.com/key4hep/EDM4hep/blob/f0d153f698c1ca3ad3c5679931b684eb296ad25e/edm4hep.yaml#L313

PR #274, opened 2024-02-22
Rename EDep to eDep in SimTrackerHit

jmcarcell:
BEGINRELEASENOTES
- Rename EDep to eDep in SimTrackerHit for consistency with other uses of "eDep"

ENDRELEASENOTES

Fixes https://github.com/key4hep/EDM4hep/issues/180
I think EDep to eDep fits better the variable usage that we have instead of eDep to EDep

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -231,7 +231,7 @@ datatypes:
     Author: "F.Gaede, DESY"
     Members:
       - uint64_t cellID      //ID of the sensor that created this hit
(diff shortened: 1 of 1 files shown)


tmadlener (commented):
This would require schema evolution, but for the generated code this is transparent as in both cases we get `getEDep` / `setEDep`, since the first letter of the member variable is capitalized.

tmadlener (approved):

PR #277, opened 2024-02-22
Use the units outside of the comments for nicer docstrings

tmadlener:
BEGINRELEASENOTES
- Use the units explictly in the yaml file to make them appear nicer in the generated docstrings.

ENDRELEASENOTES


jmcarcell:
Isn't it nice already? 

2024-02-22-180943_364x46_scrot

Where do they appear when changed like in this PR?

tmadlener:
It's basically the same (modulo the `in` IIRC). But the generator has a unit in its hand which might become handy at some point. Additionally the yaml file get's slightly easier to read because the unit is further left, compared to at the end of a comment.

PR #283, opened 2024-02-29
Adding metadata name for event filter statistics

kjvbrt:
The `vector<int>` contains three values: Seen, accepted and targeted. 

BEGINRELEASENOTES
- Adding metadata name for event filter statistics.

ENDRELEASENOTES


jmcarcell:
Which filters does this apply to? And what does targeted mean? The ones for which the filter can apply?

kjvbrt:
This is for filters which filter events e. g. : [GenEventFilter](https://github.com/HEP-FCC/k4Gen/blob/main/k4Gen/src/components/GenEventFilter.h).

The targeted number of events is number of events specified in the steering or `k4run -n`.
If someone is reading files from a input file, there might not be enough events in the file to fulfil it.
Ideally targeted and accepted should be equal.

tmadlener:
What about cases where there are multiple filters? Should there be the possibility to add a filter name?

kjvbrt:
> What about cases where there are multiple filters? Should there be the possibility to add a filter name?

Than this needs to be a vector of some struct/class (consisting of three numbers and a name for the filter)?

jmcarcell (approved):

jmcarcell:
I think having the name is fine but we can keep discussing how it's used. Also, why is the number of events requested important? How is that information going to be used? Is it useless when you request more events than available (it only tells you that there weren't enough events)

jmcarcell:
@kjvbrt any answers to my last comment? Do we want the comment to be quite restrictive? Does one always need the number of events? Maybe we should just remove the comment entirely like the other constants?

kjvbrt:
My goal was to standardize also vector members, but that is probably too restrictive.

kjvbrt (commented):
Removing comment

kjvbrt (commented):
removing comment

PR #285, opened 2024-03-12
added static asserting vector components' member ordering and padding

m-fila:
BEGINRELEASENOTES
- added static asserting vector components' member ordering and padding

ENDRELEASENOTES

The `operator[]` of generated vector classes assumes no padding and member ordering same as in the source code. These are reasonable assumptions but not required by the standard. The assumptions can be checked at a compile time

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,36 +12,36 @@ A generic event data model for future HEP collider experiments.
 
 | | | |
 |-|-|-|
-| [Vector4f](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L9)       | [Vector3f](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L27)  | [Vector3d](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L43)    |
-| [Vector2i](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L63)      | [Vector2f](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L78)  | [TrackState](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L93)  |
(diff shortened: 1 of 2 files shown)


tmadlener (approved):
I suppose we really need to check all members if we want to be sure and we don't have a shortcut of assuming that, e.g. `z` should be OK, if `x` and `y` have no padding?

Apart from the fact that our yaml files become a bit noisy, I really like this solution, and from my side I think this can be merged.

jmcarcell:
Only one nitpick: 
> The `operator[]` of generated vector classes assumes no padding and member ordering same as in the source code. These are reasonable assumptions but not required by the standard. The assumptions can be checked at a compile time

In this case the ordering is guaranteed; what we are missing to have no padding is having a default constructor. See here https://learn.microsoft.com/en-us/cpp/cpp/trivial-standard-layout-and-pod-types?view=msvc-170, although I don't know if what I say about the ordering is here but this can be found elsewhere

tmadlener:
I am not sure we need a default constructor as we generate the following:

```cpp
struct Vector3f {
  float x{};
  float y{};
  float z{};
}
```

So at least from an initialization point of view the members are default initialized.

jmcarcell:
For not having padding you need it and not to initialize the members
``` c++
struct Vector3f {
  float x{};
  float y{};
  float z{};

  Vector3f() = default;

};

struct Vector3f2 {
  float x;
  float y;
  float z;

  Vector3f2() = default;

};

int main()
{
  cout << is_pod<Vector3f>::value << endl;  // not POD
  cout << is_pod<Vector3f2>::value << endl; // POD
  return 0;
}
```

tmadlener:
I think padding and initialization are two different things here(?). `std::is_pod` is effectively `std::is_trivial && std::is_standard_layout` (IIUC). What we test for in podio is `std::is_standard_layout` as that is what we are mainly interested in. ([Corresponding unit test for the `Data` types](https://github.com/AIDASoft/podio/blob/8ce5945a878770d1d959dcf804a746281976ed67/tests/unittests/unittest.cpp#L230-L240))

PR #286, opened 2024-03-12
Make sure that operator[] return the right types

tmadlener:
BEGINRELEASENOTES
- Make sure that `operator[]` of the `VectorNx` types return the correct type.

ENDRELEASENOTES

Follow up on #285 where we somehow missed the wrong return types in the review.


jmcarcell:
Indeed this was in https://github.com/key4hep/EDM4hep/pull/285/checks#step:3:1214, which I should have checked but it always fails with EDM4hep because of some package at the end of the dependency tree

jmcarcell (approved):

tmadlener:
The downstream build seems to be back to the usual failure?

jmcarcell:
Yeah, maybe I should remove KalDet from these checks, but I guess it should compile fine with default options

PR #287, opened 2024-03-14
Add dedicated covariance matrix components and use them

tmadlener:
BEGINRELEASENOTES
- Introduce `edm4hep::CovMatrix[2,3,4,6}f` components to represent covariance matrices
  - using `std::array<float, N>` under the hood
  - Access functionality that works on enums for defining dimensions and some utilities to do the indexing into the underlying storage
- Introduce some `enum class`es for defining dimensions, e.g. `TrackParams` or `Cartesian`
- Use these components instead of the raw `std::array`s in datatypes and other components and tie the interpretation of the values to the approriate dimension `enum class`

Usage example:
```cpp
#include "edm4hep/TrackerHit3D.h"

void foo(const edm4hep::TrackerHit3D& hit) {
  const auto covXY  = hit.getCovMatrix(edm4hep::Cartesian::x, edm4hep::Cartesian:y);
}
```

ENDRELEASENOTES

I am not yet sure whether I really like the duplication of all the extra code here. Especially for the components it is effectively always the same and this could in principle be easily templated. However, then we would have to mix jinja and c++ templates, so I am not entirely sure if we want that.

One possibility would be to put that part of the extra code into some `ipp` file and just `#include` it in the ExtraCode declaration.

- [ ] Check breakages downstream
  -  https://github.com/HEP-FCC/FCCAnalyses/pull/362 to build the stack with and without these changes
  - https://github.com/key4hep/k4EDM4hep2LcioConv/pull/56 in order to not break the tests there

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -132,22 +132,8 @@ components:
           constexpr CovMatrix2f& operator=(std::array<float, 3> v) { values = v; return *this; }\n
           bool operator==(const CovMatrix2f& v) const { return v.values == values; }\n
           bool operator!=(const CovMatrix2f& v) const { return v.values != values; }\n
-          constexpr float operator[](unsigned i) const { return values[i]; }\n
-          constexpr float& operator[](unsigned i) { return values[i]; }\n
-          constexpr auto begin() const { return values.begin(); }\n
-          constexpr auto begin() { return values.begin(); }\n
-          constexpr auto end() const { return values.begin(); }\n
-          constexpr auto end() { return values.begin(); }\n
-          float* data() { return values.data(); }\n
-          const float* data() const { return values.data(); }\n
-          /// Get the value of the covariance matrix for the two passed dimensions\n
-          template<typename DimEnum>\n
-          constexpr float getValue(DimEnum dimI, DimEnum dimJ) const { return utils::get_cov_value(values, dimI, dimJ); }\n
-          /// Set a value in the covariance matrix for the two passed dimensions\n
-          template<typename DimEnum>\n
-          constexpr void setValue(float value, DimEnum dimI, DimEnum dimJ) { utils::set_cov_value(value, values, dimI, dimJ); }\n
-        "
-
+          #include <edm4hep/detail/CovMatrixCommon.ipp>
+          "
     edm4hep::CovMatrix3f:
       Description: "A generic 3 dimensional covariance matrix with values stored in lower triangular form"
       Members:
@@ -160,20 +146,7 @@ components:
           constexpr CovMatrix3f& operator=(std::array<float, 6> v) { values = v; return *this; }\n
           bool operator==(const CovMatrix3f& v) const { return v.values == values; }\n
           bool operator!=(const CovMatrix3f& v) const { return v.values != values; }\n
-          constexpr float operator[](unsigned i) const { return values[i]; }\n
-          constexpr float& operator[](unsigned i) { return values[i]; }\n
-          constexpr auto begin() const { return values.begin(); }\n
-          constexpr auto begin() { return values.begin(); }\n
-          constexpr auto end() const { return values.begin(); }\n
-          constexpr auto end() { return values.begin(); }\n
-          float* data() { return values.data(); }\n
-          const float* data() const { return values.data(); }\n
-          /// Get the value of the covariance matrix for the two passed dimensions\n
-          template<typename DimEnum>\n
-          constexpr float getValue(DimEnum dimI, DimEnum dimJ) const { return utils::get_cov_value(values, dimI, dimJ); }\n
-          /// Set a value in the covariance matrix for the two passed dimensions\n
-          template<typename DimEnum>\n
-          constexpr void setValue(float value, DimEnum dimI, DimEnum dimJ) { utils::set_cov_value(value, values, dimI, dimJ); }\n
+          #include <edm4hep/detail/CovMatrixCommon.ipp>
         "
 
     edm4hep::CovMatrix4f:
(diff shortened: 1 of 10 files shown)


tmadlener:
The release based CI workflows fail because https://github.com/AIDASoft/podio/pull/553 is not yet in there.

edm4hep.yaml
```diff
@@ -464,7 +631,7 @@ datatypes:
       - float                  charge         // charge of the reconstructed particle
       - float                  mass  [GeV]    //  mass of the reconstructed particle, set independently from four vector. Four momentum state is not kept consistent internally
       - float                  goodnessOfPID  // overall goodness of the PID on a scale of [0;1]
-      - std::array<float,10>   covMatrix      // cvariance matrix of the reconstructed particle 4vector (10 parameters). Stored as lower triangle matrix of the four momentum (px,py,pz,E), i.e. cov(px,px), cov(py,##
+      - edm4hep::CovMatrix4f   covMatrix      // cvariance matrix of the reconstructed particle 4vector (10 parameters). Stored as lower triangle matrix of the four momentum (px,py,pz,E), i.e. cov(px,px), cov(py,##
```
m-fila, line 634: typo, close parenthesis to silence overeager editors
```suggestion
      - edm4hep::CovMatrix4f   covMatrix      // covariance matrix of the reconstructed particle 4vector (10 parameters). Stored as lower triangle matrix of the four momentum (px,py,pz,E), i.e. cov(px,px), cov(py,##)
```

utils/include/edm4hep/utils/cov_matrix_utils.h
```diff
@@ -0,0 +1,119 @@
...
+#include <iostream>
+#include <stdexcept>
+#include <type_traits>
+
+namespace edm4hep {
+
+namespace utils {
+
+  namespace detail {
+// From c++23 this is functionality offerd by the STL
```
m-fila, line 16: typo
```suggestion
// From c++23 this is functionality offered by the STL
```

tmadlener:
I have also removed the storage details from the docstrings of the covariance matrix members, as that should be (almost) completely be abstracted away by the components.

tmadlener:
Because this came up at the [EDM4hep meeting](https://indico.cern.ch/event/1409361/?note=277156): This still has the same issues for `RDataFrame::Describe` as the version with bare `std::array`. Trying to run it on the example file that gets created during the tests, I get:

```
TClass::Init:0: RuntimeWarning: no dictionary for class edm4hep::CovMatrix3f is available
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert edm4hep::TrackerHitPlaneData::covMatrix from type: edm4hep::CovMatrix3f to type: array<float,6>, skip element
---------------------------------------------------------------------------
runtime_error                             Traceback (most recent call last)
Cell In[3], line 1
----> 1 df.Describe()

runtime_error: ROOT::RDF::RDFDescription ROOT::RDF::RInterfaceBase::Describe() =>
    runtime_error: TTree leaf TrackerHitPlanes.covMatrix.values[6] has both a leaf count and a static length. This is not supported.

```

edm4hep.yaml
```diff
@@ -357,17 +422,30 @@ datatypes:
...
+      declaration: "
+      /// Get the position error value for the two passed dimensions\n
+      float getPositionError(edm4hep::Cartesian dimI, edm4hep::Cartesian dimJ) const { return getPositionError().getValue(dimI, dimJ); }\n
+      "
+    MutableExtraCode:
+      includes: "#include <edm4hep/Constants.h>"
+      declaration: "
+      /// Set the position error value for the two passed dimensions\n
+      void setPositionError(float value, edm4hep::Cartesian dimI, edm4hep::Cartesian dimJ) { return getPositionError().setValue(value, dimI, dimJ); }\n
+      void setPositionError(std::array<float, 6> values) { getPositionError() = values; }
```
andresailer, line 447:
```suggestion
      void setPositionError(std::array<float, 6> values) { positionError = values; }
```

Doesn't getPositionalError() return `const std::array&` ? Similar in almost all other setters introduced in this PR.

tmadlener: It will return a const-correct reference to a `std::array`. This has been introduced with https://github.com/AIDASoft/podio/pull/553

andresailer: OK. [EDM4hep Doxygen](https://edm4hep.web.cern.ch/classedm4hep_1_1_reconstructed_particle.html#a1f6810a43c73575dba7abc54e780db79) hasn't caught up yet I think 

Other thing
```suggestion
      void setPositionError(const std::array<float, 6>& values) { getPositionError() = values; }
```
as far as I understand the assignment operator sets item by item, so this would otherwise make a copy of the input array.
https://en.cppreference.com/w/cpp/container/array
> overwrites every element of the array with the corresponding element of another array

tmadlener: > OK. [EDM4hep Doxygen](https://edm4hep.web.cern.ch/classedm4hep_1_1_reconstructed_particle.html#a1f6810a43c73575dba7abc54e780db79) hasn't caught up yet I think

It has; However only the [Mutable types](https://edm4hep.web.cern.ch/classedm4hep_1_1_mutable_reconstructed_particle.html#a7009b2d2db2fe114ab181fe8edea7fc7) get the mutable reference (and the corresponding `const` overload).

> as far as I understand the assignment operator sets item by item, so this would otherwise make a copy of the input array.

Isn't the effect the same in the end? What we want is to have the values that we pass in to be set inside our data. Whether that is done by one copy or by several individual assignments shouldn't matter too much?

tmadlener: At least with `-O2` it doesn't seem to make a big difference: [compiler explorer](https://godbolt.org/z/GfWqrKh14). IIUC this both boils down to effectively copying over numbers in chunks of whatever vector register size is available on the underlying hardware.

I also quickly checked with `-O0` as well and the biggest difference seems to be the initial handling of `const&` vs passing by value.

andresailer: So is there any argument against using `const&` ?

tmadlener: No, I don't think so.

edm4hep.yaml
```diff
@@ -141,6 +201,11 @@ components:
         static const int AtCalorimeter = 4 ;\n
         static const int AtVertex = 5 ;\n
         static const int LastLocation = AtVertex  ;\n
+
+        /// Get the covariance matrix value for the two passed parameters\n
+        constexpr float getCovMatrix(edm4hep::TrackParams parI, edm4hep::TrackParams parJ) const { return covMatrix.getValue(parI, parJ); }\n
+        /// Set the covariance matrix value for the two passed parameters\n
+        constexpr void setCovMatrix(float value, edm4hep::TrackParams parI, edm4hep::TrackParams parJ) { covMatrix.setValue(value, parI, parJ); }
```
andresailer, line 208: setCovMatrix(std::array const&) missing

m-fila: Shouldn't the `setCovMatrix` be put in `MutableExtraCode` instead of `ExtraCode`?

tmadlener: Definitely. Let me check again to have that consistently.

tmadlener: After having had another look; The `TrackState` is actually a component. Hence, it only has `ExtraCode` and it doesn't need a `setCovMatrix` because you can just assign to `covMatrix` directly (without having to go through the setter).

tmadlener:
I have removed the `setCovMatrix(const std::array<float, N>&)` from the datatypes and instead added a constructor of the form

```cpp
template <typename ...Vs>
CovMatrixNf(Vs... vs) : values{vs...} {}
```

to the covariance matrix components. This makes direct initialization from initializer lists via the datatypes possible, e.g. as in the tests

```cpp
  auto trackerHit = MutableTrackerHit3D{};
  trackerHit.setCovMatrix({1.f, 2.f, 3.f, 4.f, 5.f, 6.f});
  REQUIRE(trackerHit.getCovMatrix() == std::array{1.f, 2.f, 3.f, 4.f, 5.f, 6.f});

  const std::array arrValues = {6.f, 5.f, 4.f, 3.f, 2.f, 1.f};
  trackerHit.setCovMatrix(arrValues);
  REQUIRE(trackerHit.getCovMatrix() == std::array{6.f, 5.f, 4.f, 3.f, 2.f, 1.f});
```

PR #299, opened 2024-05-07
WIP: Add Reconstructed TOF.

mirguest:
BEGINRELEASENOTES
- Add Reconstructed TOF info
ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -781,6 +781,19 @@ datatypes:
     OneToOneRelations:
       - edm4hep::Track track           // the corresponding track
 
+  #-------------  RecTof
+  edm4hep::RecTof:
+    Description: "Reconstructed TOF (time of flight) info."
+    Author: "EDM4hep authors"
+    Members:
+      - float time [ns]                      // time measurement
+      - std::array<float, 5> timeExp [ns]    // expected time for e(0),mu(1),pi(2),K(3),p(4)
+      - float sigma                          // time resolution
+      - std::array<float, 5> pathLength [mm] // length of flight for e(0),mu(1),pi(2),K(3),p(4)
+      - edm4hep::Vector3d position [mm]      // extrapolated hit position
+    OneToOneRelations:
+      - edm4hep::Track track                 // the corresponding track
+
 interfaces:
   edm4hep::TrackerHit:
     Description: "Tracker hit interface class"


edm4hep.yaml
```diff
@@ -781,6 +781,19 @@ datatypes:
...
+  #-------------  RecTof
+  edm4hep::RecTof:
+    Description: "Reconstructed TOF (time of flight) info."
+    Author: "EDM4hep authors"
+    Members:
+      - float time [ns]                      // time measurement
+      - std::array<float, 5> timeExp [ns]    // expected time for e(0),mu(1),pi(2),K(3),p(4)
+      - float sigma                          // time resolution
+      - std::array<float, 5> pathLength [mm] // length of flight for e(0),mu(1),pi(2),K(3),p(4)
+      - edm4hep::Vector3d position [mm]      // extrapolated hit position
```
tmadlener, line 793: I would potentially re-order these to keep the particle species bits and the "measured" bits together, e.g.
- `time`
- `sigma`
- `position`

`sigma` still needs a unit, which, I would assume is `ps`?

Is the extrapolated hit position necessary here? I assume you need it for the calculation of the track length, but do you still need it afterwards?

tmadlener (commented):
Quick summary from the discussion that happened at the [EDM4hep meeting](https://indico.cern.ch/event/1413652/) (see the attached notes).

> * Do we need the 5 different path lengths and expected times for the 5 particle species?
>    * Different track length / time for different species
>   * Could create new tracks for the different species and link to that
> * Refitting (and storing) the track might be prohibitive due to size (and information duplication)
>    * No clear best solution to this yet. Needs some further discussion

At this point in time it's not yet clear whether refitting **and storing** tracks for each particle species and then having a simpler `RecTof` is "better" than the current design, where there is the one *canonical track* that can then be linked to the current (more complicated) `RecTof`.

Some minor questions / comments also below.

edm4hep.yaml
andresailer: I would also propose `sigma` -> `timeError`, which is similar to other places where we have value and valueError

m-fila: How about renaming `sigma` to something like `resolution`

andresailer: https://github.com/search?q=repo%3Akey4hep%2FEDM4hep+error+language%3AYAML&type=code&l=YAML

edm4hep.yaml
```diff
@@ -781,6 +781,19 @@ datatypes:
     OneToOneRelations:
       - edm4hep::Track track           // the corresponding track
 
+  #-------------  RecTof
+  edm4hep::RecTof:
+    Description: "Reconstructed TOF (time of flight) info."
+    Author: "EDM4hep authors"
+    Members:
+      - float time [ns]                      // time measurement
+      - std::array<float, 5> timeExp [ns]    // expected time for e(0),mu(1),pi(2),K(3),p(4)
```
tmadlener, line 790: Should we make this a `VectorMember` to have a flexible number of hypotheses, instead of limiting ourselves to 5?

edm4hep.yaml
```diff
@@ -781,6 +781,19 @@ datatypes:
...
+      - float sigma                          // time resolution
+      - std::array<float, 5> pathLength [mm] // length of flight for e(0),mu(1),pi(2),K(3),p(4)
```
andresailer, line 792: Add a "unknown/background/other" entry?

edm4hep.yaml
andresailer: The consensus from the discussion in the Key4hep meeting is that this is too much freedom.

PR #302, opened 2024-05-08
Update the schema diagram to reflect TrackerHit as interface

tmadlener:
BEGINRELEASENOTES
- Update the schema diagram to reflect the `TrackerHit` as interface
  - Add `TrackerHit3D` as new type
- Fix all arrow heads to be consistently black

ENDRELEASENOTES

### DIFF
--- a/doc/edm4hep_diagram.svg
+++ b/doc/edm4hep_diagram.svg
@@ -24,12 +24,12 @@
      inkscape:pageopacity="0"
      inkscape:pageshadow="2"
      inkscape:window-width="1920"
-     inkscape:window-height="1016"
+     inkscape:window-height="1136"
      id="namedview332"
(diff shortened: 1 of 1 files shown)


jmcarcell (approved):
I'm not fully convinced by how the interface is displayed, doesn't look very clear that the arrows from `TrackerHit` to `TrackerHitPlane` or `TrackerHit3D` are different from the other arrows in the diagram. But I don't have any suggestions with the space that is available, so :+1:

tmadlener:
Yeah, I am also not yet entirely convinced it's the best solution, but it is kind of close to what was there before and I didn't have to rework the full diagram ;)

jmcarcell:
No one has said anything so I think this can go in

PR #303, opened 2024-05-08
Update the schema diagram to reflect the new ParticleID to ReconstructedParticle relation

tmadlener:
BEGINRELEASENOTES
- Update the schema diagram to reflect the changes from [#268](https://github.com/key4hep/EDM4hep/pull/268)

ENDRELEASENOTES

- [x] Builds on #302 

I split this into two PRs, because #302 shouldn't be too controversial, while removing the version might be a bit less clear

### DIFF
--- a/doc/edm4hep_diagram.svg
+++ b/doc/edm4hep_diagram.svg
@@ -24,11 +24,11 @@
      inkscape:pageopacity="0"
      inkscape:pageshadow="2"
      inkscape:window-width="1920"
-     inkscape:window-height="1136"
+     inkscape:window-height="1016"
      id="namedview332"
      showgrid="true"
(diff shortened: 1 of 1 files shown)


jmcarcell (approved):
Needs https://github.com/key4hep/EDM4hep/pull/304 to be correct.
I don't mind removing the version, but is it permanent? Otherwise since it doesn't seem there are going to be so many changes before the next version we could just tag it with 1.0 or 0.10.99 (a bit ugly) or maybe pre-1.0

andresailer:
Keeping the version would be nice indeed.

tmadlener:
Brought the version back, but I am not entirely sure what to set it to for the moment.

PR #305, opened 2024-05-16
Transparently introduce switch to optional param values

tmadlener:
BEGINRELEASENOTES
- Introduce pre-processor checks to transparently switch to the new `std::optional` return values of `podio::Frame::getParameter` (introduced with [AIDASoft/podio#580](https://github.com/AIDASoft/podio/pull/580))

ENDRELEASENOTES

### DIFF
--- a/test/read_events.h
+++ b/test/read_events.h
@@ -11,12 +11,10 @@
 
 // podio specific includes
 #include "podio/Frame.h"
+#include "podio/podioVersion.h"
 
 // STL
-#include <cassert>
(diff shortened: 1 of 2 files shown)


hegner:
@tmadlener - I think we should at the same time open a ticket to drop this after the 1.0 version is out. Do not like to accumulate these things

tmadlener:
Since this is mainly concerning one test, and the not really far spread particle id utils, I will merge this shortly, unless someone complains.

PR #307, opened 2024-05-16
Use MurmurHash3 to hash the algorithm name for the algorithm type in ParticleIDMeta

### LINKED ISSUE #298: Stable / consistent algorithm types for PIDHandler
The PIDHandler introduced with #268 leaves the determination / setting of the `algorithmType` / `algoID` value to the user. This is quite error prone and we should make it easy / trivial to set a consistent & unique value here via the same tools. The easiest solution is probably to use some hash of the algorithm name.

This issue is mainly here for keeping track of this, as it was raised during the review of #268, but not yet implemented there.

tmadlener:
BEGINRELEASENOTES
- To have a more consistent way of filling the `algoType` in the `edm4hep::utils::ParticleIDMeta` use the 32 bit version of MurmurHash3 to hash the `algoName` to get to `algoType`. Fixes #298 
  - Make `algoType` a private member but allow read access to it via the `algoType` member function. `algoType` has to be filled on construction. It is still possible to set it manually, the hashing will only kick in for the constructor taking only a name.

ENDRELEASENOTES

We could consider making `ParticleID::algoType` a `uint32_t` field for this, as that would make it a bit more easily visible that this is potentially a hash.


jmcarcell:
Copying the murmurhash files every time we want to hash something isn't great...

tmadlener:
I agree, but I am not sure where to put them for EDM4hep, and I don't really want to expose them from podio.

jmcarcell:
A possibility could be to make a new Murmurhash package and build that library. Can any other hashes be considered, like `boost::hash` or is that a no no? The hashes here are saved in files as metadata right?

utils/src/MurmurHash3.cpp
```diff
@@ -0,0 +1,442 @@
+//-----------------------------------------------------------------------------
+// MurmurHash3 was written by Austin Appleby, and is placed in the public
+// domain. The author hereby disclaims copyright to this source code.
+
+// Note - The x86 and x64 versions do _not_ produce the same results, as the
```
andresailer, line 5: Does the arm64 version a) exist and b) produce the same result as for x64?

tmadlener:
I don't have a strict preference for any hashing algorithm. In the end the main requirement is that it has to fit into 32 bits. Whether we want to pull in boost just for that is another question though.

utils/src/MurmurHash3.cpp
tmadlener: Good question, I don't know. Maybe @hegner can say something on that?

andresailer: Probably doesn't matter, because only `MurmurHash3_x86_32` is used. Which means less than optimal performance, but probably doesn't matter either.

utils/src/ParticleIDUtils.cc
```diff
@@ -11,6 +13,22 @@
 
 namespace edm4hep::utils {
 
+namespace {
+  int32_t getAlgoID(const std::string& name) {
+    int32_t ID = 0;
+    MurmurHash3_x86_32(name.c_str(), name.size(), 0, &ID);
```
andresailer, line 19: If we only use this hash implementation, we could drop the other unused implementations.

tmadlener:
Unless there are any more comments, I would merge this sometime tomorrow (synchronously with the necessary API changes in https://github.com/key4hep/k4EDM4hep2LcioConv/pull/68)

PR #308, opened 2024-05-17
Use 'declarationFile' instead of  #include in CovMatrix component

### LINKED ISSUE #296: Missing methods in doxygen documentation for `CovMatrix` classes
In `CovMatrix` classes the members declared with `#include <edm4hep/detail/CovMatrixCommon.ipp>` don't appear in documentation generated with doxygen. For instance `operator[]`, `getValue` and `setValue` are missing. 

Doxygen is doing some limited preprocessing and apparently doesn't resolve correctly `#include` directives inside `{ }`.

https://github.com/key4hep/EDM4hep/blob/f46cf67f39d5d4b587e08359786d808a1c5c8ea3/edm4hep.yaml#L130-L138

missing methods

m-fila:
BEGINRELEASENOTES
- Use declarationFile instead of include in CovMatrix to fix generated documentation for covariance matrix components. Fixes [#296](https://github.com/key4hep/EDM4hep/issues/296)

ENDRELEASENOTES

Closes #296 

Declarations added with `#include` were not appearing in a documentation generated with doxygen. New podio direcative 'declarationFile'  (https://github.com/AIDASoft/podio/pull/601)  includes  these declaration during file generation without relaying on pre-processor and make them visible to doxygen
| Old | New |
|-|-|
|Screenshot_2024-05-01_14-27-21|covmatrix_fix |

### DIFF
--- a/edm4hep/extra_code/CovMatrixCommon.ipp
+++ b/edm4hep/extra_code/CovMatrixCommon.ipp
@@ -1,5 +1,5 @@
-// This file is meant to be included inside the declaration part of the
-// ExtraCode for the CovMatrixNx components. They live in this file because they
+// This file is meant to be included with ExtraCode  declarationFile directive
+// for the CovMatrixNx components. They live in this file because they
 // can be written very generically and reduce the clutter and code repetition in
 // the edm4hep.yaml file
 //
--- a/edm4hep/extra_code/CovMatrixCommon.ipp
+++ b/edm4hep/extra_code/CovMatrixCommon.ipp
@@ -1,4 +1,4 @@
-// This file is meant to be included with ExtraCode  declarationFile directive
+// This file is meant to be included via the ExtraCode declarationFile directive
 // for the CovMatrixNx components. They live in this file because they
 // can be written very generically and reduce the clutter and code repetition in
 // the edm4hep.yaml file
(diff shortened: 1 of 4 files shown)


edm4hep/extra_code/CovMatrixCommon.ipp
```diff
@@ -1,5 +1,5 @@
-// This file is meant to be included inside the declaration part of the
-// ExtraCode for the CovMatrixNx components. They live in this file because they
+// This file is meant to be included with ExtraCode  declarationFile directive
```
tmadlener, line 1:
```suggestion
// This file is meant to be included via the ExtraCode declarationFile directive
```

tmadlener:
I think we also need to install the `extra_code` directory into `<prefix>/share` (or wherever we put the yaml file exactly). Otherwise downstream models cannot use EDM4hep as an upstream. See, e.g. https://github.com/key4hep/EDM4hep/actions/runs/9205295127/job/25320679012?pr=308#step:4:1220

m-fila:
The `extra_code` is installed in `<prefix>/share/edm4hep/edm4hep/extra_code` to mimic the path in source tree. Or in the source tree it could be move  from `edm4hep/extra_code` to the top level and then installed in `<prefix>/share/edm4hep/extra_code`

tmadlener (approved):

hegner:
@m-fila  this PR destroyed all automatic testing because the keyword `DEPENDS` is interpreted as value of ` PODIO_IO_HANDLERS`. Can you prepare a fix?
ping @tmadlener

tmadlener:
As far as I can tell only those workflows which build on an old(ish) version of podio are currently broken. However, there are multiple small changes in EDM4hep that depend on features of podio that came after v00.99

m-fila:
this should break only  the `release` checks. Do you mean there are some other checks failing or should I fix the `release` checks? If the second is the case I don't really see any solutions other than reverting this

tmadlener:
I don't think it's worth fixing the release builds. They effectively just have a podio version that is no longer sufficient. We could put a version requirement into the top level CMakeLists.txt if we wanted to, to make the tests fail earlier. However, since podio v1.0 is very close, I think we just wait with that until we have that.

Issue #309, opened 2024-05-21
Add explicit storage types for Generator information


hegner:
As discussed on May 21 provide explicit storage types for MC information.

See issue https://github.com/key4hep/EDM4hep/issues/208 

- String values may have to stay as parameters
- Rethink names
- Provide an explicit structure containing the per-event information
- Provide an explicit structure for the `ToolInfo`
- "Optional" information as much as possible in MCParticle as not to spread information over too many collections. Probably no change to it needed. Color flow etc should be fine there. Optional values can be implemented as storing "0".
- Store kinematics information in double rather than float

PR #310, opened 2024-05-22
Add datatypes for storing generator (meta)data in a structured and defined way

hegner:
BEGINRELEASENOTES
- Add `GeneratorEventParameters` and `GeneratorPdfInfo` datatypes to store generator related data in a structured and well defined way.
- Add a `GeneratorToolInfo` struct and related utility functionality to store some high level metadata on the generator into Frame parameters.

ENDRELEASENOTES

Pull request to collect all the changes required for #309


edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+  #---------- GeneratorParameters
+  edm4hep::GeneratorParameters:
+    Description: "Generator parameters"
+    Author: "EDM4hep authors"
+    Members:
+      - double event_scale              // event scale
+      - double alphaQED                 // alpha_QED
+      - double alphaQCD                 // alpha_QCD
+      - int signal_process_id           // id of signal process
+      - double sqrt_s                   // sqrt(s)
```
tmadlener, line 795:
```suggestion
      - double sqrt_s  [GeV]            // sqrt(s)
```

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,48 @@
+#ifndef EDM4HEP_GENTOOLINFO_H
```
tmadlener, line 1: Should this go to the `utils`? Mainly because, e.g. the ParticleID helpers also live there.

hegner: it could. The struct however I'd like to keep higher level.

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,47 @@
...
+  std::string name;
+  std::string version;
+  std::string description;
+};
+
+const std::vector<GenToolInfo> getGenToolInfos(podio::Frame& frame) {
+  auto toolInfos = std::vector<GenToolInfo>();
+  auto names = frame.getParameter<std::string>("name");
+  auto versions = frame.getParameter<std::string>("version");
+  auto descriptions = frame.getParameter<std::string>("description");
```
tmadlener, line 21: I think the parameter names / keys need to a bit more descriptive here. Otherwise the chances for clashes is quite high. We should also put them into our `Constants.h` header so that there is a sort of canonical form to refer to them.

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,47 @@
...
+  for (unsigned int i = 0; i < names.size(); ++i) {
+    auto toolInfo = GenToolInfo();
+    toolInfo.name = names[i];
+    toolInfo.version = versions[i];
+    toolInfo.description = descriptions[i];
```
tmadlener, line 26: We are missing at least a `push_back` here. Since we have all the infos already we can also just do an `emplace_back` here, I think.
```suggestion
    toolInfo.emplace_back(names[i], versions[i], descriptions[i]);
```

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,47 @@
...
+  auto versions = std::vector<std::string>();
+  auto descriptions = std::vector<std::string>();
+  for (auto& toolInfo : toolInfos){
+    names.push_back(toolInfo.name);
+    versions.push_back(toolInfo.version);
+    descriptions.push_back(toolInfo.description);
+  }
+  frame.putParameter("name", names);
+  frame.putParameter("version", versions);
+  frame.putParameter("description", descriptions);
```
tmadlener, line 42:
```suggestion
  frame.putParameter("name", std::move(names));
  frame.putParameter("version", std::move(versions));
  frame.putParameter("description", std::move(descriptions));
```
We can avoid an unnecessary copy here even if it shouldn't matter too much, I think.

tmadlener (commented):
I think overall this looks like a reasonable structure (at least to me). But @dirkzerwas should probably have a look to make sure we don't miss anything obvious yet.

Can you also add tests to read back the things that you currently write. At least the ToolInfo parts have enough additional logic wrt podio to warrant a test, I think.

include/edm4hep/GenToolInfo.h
hegner: yap. that's a good point. I was considering putting that into the constants indeed. Next iteration will have it

dirkzerwas:
Hi @hegner and @tmadlener,

looks very nice, a couple of thoughts:
- just wondering if adding ''Event" into GeneratorParameters might make sense to make clear it's the "per event" structure and not the general setting of the generator (like the N PYTHIA switches)
- I may have missed it: HepMC has a vector of weightnames in the run_info part, suggest to foresee that in "run" (even though this is not filled by all generators consistently at the moment)
- weights: do you wan to leave the weights in the EventHeader? In principle one could argue that weights are linked with the cross section @apricePhy told me that the use is not the some in all generators. So to be safe, one may want to keep the two together?

Cheers,
Dirk

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+    Description: "Generator pdf information"
+    Author: "EDM4hep authors"
+    Members:
+      - std::array<int, 2> parton_id      // Parton PDG id
+      - std::array<int, 2> id             // LHAPDF PDF id (see https://lhapdf.hepforge.org/pdfsets.html)
+      - std::array<double, 2> x           // Parton momentum fraction
+      - std::array<double, 2> xf          // PDF value
+      - double scale [GeV]                // Factorisation scale
+    OneToManyRelations:
+      - edm4hep::MCParticle signal_vertex // pointing into MC Generators collection
```
andresailer, line 811:
```suggestion
      - edm4hep::MCParticle signal_vertex // pointing into MCParticle collection
```
? 
Or what is a MC Generators collection?

And what does this mean anyway? Because it being a relation already means that it points to a MCParticle collection. So what does `signal_vertex` actually mean?

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,44 @@
...
+  std::string description;
+};
+
+const std::vector<GenToolInfo> getGenToolInfos(podio::Frame& frame) {
+  auto toolInfos = std::vector<GenToolInfo>();
+  auto names = frame.getParameter<std::string>("name");
+  auto versions = frame.getParameter<std::string>("version");
+  auto descriptions = frame.getParameter<std::string>("description");
+  for (unsigned int i = 0; i < names.size(); ++i) {
+    toolInfo.emplace_back(names[i], versions[i], descriptions[i]);
```
jmcarcell, line 23:
```suggestion
    toolInfos.emplace_back(names[i], versions[i], descriptions[i]);
```
missing s (or extra s above)

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,44 @@
+#ifndef EDM4HEP_GENTOOLINFO_H
+#define EDM4HEP_GENTOOLINFO_H
+
+#include "podio/Frame.h"
+#include "Constants.h"
```
andresailer, line 5:
```suggestion
#include "edm4hep/Constants.h"
```

edm4hep.yaml
dirkzerwas: Hi @andresailer,
since HepMC separates particles and vertices, they can identify the signal vertex by a a number. The use is to separate the hard scatter vertex (signal_vertex) to differentiate from overlaid pileup events, so even after merging one has a chance to separate signal from background/. 
So I guess it is used as a list of all particles belonging to this particular vertex?

Wondering (I had overlooked that it is part of pdfInfo): shouldn't this list/onetomany be part of the GeneratorParameters (rather than PDFINFO)?

Cheers,
Dirk

andresailer: So this would be the `List of MCParticles that were generated by this Generator`?

dirkzerwas: sorry, I had not checked the HepMC definition, so "belonging" is very vague:
from the HepMC documentation: "V vertex OID status (comma-separated list of incoming OIDs)."
so in that spirit I would write it as the list of incoming particles of this particular vertex.

hegner:
> * just wondering if adding ''Event" into GeneratorParameters might make sense to make clear it's the "per event" structure and not the general setting of the generator (like the N PYTHIA switches)
Yes. I think that's a fair point. I will do that rename.

> * I may have missed it: HepMC has a vector of weightnames in the run_info part, suggest to foresee that in "run" (even though this is not filled by all generators consistently at the moment)
For this we do not need a new data type at all. My idea was to just have an example for how to do it in a to-be-written doc page. With possibly a constant defined for the name of the weight names. 

> * weights: do you wan to leave the weights in the EventHeader? In principle one could argue that weights are linked with the cross section @apricePhy told me that the use is not the some in all generators. So to be safe, one may want to keep the two together?
> 
I left it in one piece for simplicity. No problem to split it in two.

hegner:
@dirkzerwas - I have a question for clarification. We have event weights stored in our standard `edm4hep::EventHeader` and I would not like to change its definition. So we could either leave the cross sections in the `GeneratorEventParameters` or make it an independent type. What do you think?

dirkzerwas:
> @dirkzerwas - I have a question for clarification. We have event weights stored in our standard `edm4hep::EventHeader` and I would not like to change its definition. So we could either leave the cross sections in the `GeneratorEventParameters` or make it an independent type. What do you think?

If downstream algs already access the weights, I understand the difficulty, it was more esthetics than "need" :)
- the comment came initially at a conversation with @tmadlener (when discussing weights in general a couple of weeks ago), that by adding "weights" to EventHeader this mixes MC and "in data available" information, correct Thomas?
- I would prefer to keep the crossSections where you have put them and not extract them, the structure is clearer

However I would move the signal_vertex (oneToManyRelation) out of the PDF structure if possible?

edm4hep.yaml
andresailer: @hegner We definitely need a better docstring for this line

@dirkzerwas We don't have vertices in the HepMC3 sense. So can you give an example of what signal_vertex is for some example event?

hegner:
> However I would move the signal_vertex (oneToManyRelation) out of the PDF structure if possible?
@dirkzerwas - no problem. It would then be a standalone collection. I was hoping to keep the number of collections small however. If that's your only comment, I will implement that and finish this feature with writing a simple example.

edm4hep.yaml
hegner: @andresailer - yes. Indeed. We'll wait for @dirkzerwas to have a good example so that we write the proper thing

tmadlener:
> the comment came initially at a conversation with @tmadlener (when discussing weights in general a couple of weeks ago), that by adding "weights" to EventHeader this mixes MC and "in data available" information, correct Thomas?

Yes, exactly. In principle an EventHeader should only only contain things that are available on both simulation and real data. From a "historical point of view" we can only store multiple weights in there since #254, so there is not too much historical baggage, and this has not yet been part of a released version of EDM4hep. We will probably not get rid of the single weight in the EventHeader as easily, as that has been there since the beginning. So we could in principle move the weights vector closer to the cross sections without boo much breakage.

dirkzerwas:
@hegner
I would move the 
OneToManyRelations:
edm4hep::MCParticle signal_vertex // pointing into MCParticle collection
to GeneratorEventParameters which already has event_scale etc
but maybe I misunderstood your comment?

hegner:
@dirkzerwas - perfect

edm4hep.yaml
dirkzerwas: @andresailer 
here is an example from SHERPA. Process ZH@250GeV (I removed some stuff
HepMC::Version 3.02.07
HepMC::Asciiv3-START_EVENT_LISTING
T SHERPA-MC\|2.2.15\|Cho Oyu
E 0 7 16
U GEV MM
W 2.3990527749323822814809e-01 2.9672764031814940783455e-04 1.2613634180205666179830e+00 1.0000000000000000000000e+00 0.0000000000000000000000e+00
A 0 GenCrossSection 2.39905277e-01 2.39905277e-01 -1 -1
A 0 GenPdfInfo 11 -11 9.99988759e-01 9.22321388e-01 2.40092563e+02 1.13670760e+04 6.07635210e-01 0 0
A 0 alphaQCD 0.103428260737051
A 0 alphaQED 0.0077638411389137
A 0 event_scale 57644.4387826361
A 0 mpi -1
A 0 signal_process_vertex -5
...
P 1 0 11 0.000000000000e+00 0.000000000000e+00 1.250000000000e+02 1.250000000000e+02 0.000000000000e+00 4
P 2 1 11 0.000000000000e+00 0.000000000000e+00 1.250000000000e+02 1.250000000000e+02 0.000000000000e+00 11
P 3 0 -11 0.000000000000e+00 0.000000000000e+00 -1.250000000000e+02 1.250000000000e+02 0.000000000000e+00 4
P 4 3 -11 0.000000000000e+00 0.000000000000e+00 -1.250000000000e+02 1.250000000000e+02 0.000000000000e+00 11
P 5 4 -11 -0.000000000000e+00 -0.000000000000e+00 -1.152901735522e+02 1.152901735533e+02 5.110039371923e-04 11
P 6 4 22 0.000000000000e+00 0.000000000000e+00 -9.709826447783e+00 9.709826446650e+00 -1.482971707319e-04 1
P 7 2 11 -0.000000000000e+00 -0.000000000000e+00 1.249985948618e+02 1.249985948628e+02 5.110003775412e-04 11
P 8 2 22 0.000000000000e+00 0.000000000000e+00 1.405138223774e-03 1.405137179276e-03 -1.713279468285e-06 1
V -6 4 [5,7]
P 9 -6 11 0.000000000000e+00 0.000000000000e+00 1.249985948629e+02 1.249985948629e+02 1.907348632812e-06 3
P 10 -6 -11 0.000000000000e+00 0.000000000000e+00 -1.152901735533e+02 1.152901735533e+02 -2.336015455993e-06 3
V -5 1 [9,10]
...
so signal_vertex, as I read it, the important Vertex is "-5" V -5 refers to P 9 and 10 as incoming particles (electron, positron, but after ISR, so it is not the beam particles).
So if we want to keep the capability of moving seamless between HepMC and EDM4HEP, I would write the P9 and P10 particles into the OneToManyRelation.

Come to think of it: if we calculate the invariant mass of P9+P10 we get the stored event_scale. So I think it's a consistent definition (implemented by Sherpa at least).

dirkzerwas: sounds good to me!

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+    Description: "Generator pdf information"
+    Author: "EDM4hep authors"
+    Members:
+      - std::array<int, 2> parton_id      // Parton PDG id
+      - std::array<int, 2> id             // LHAPDF PDF id (see https://lhapdf.hepforge.org/pdfsets.html)
+      - std::array<double, 2> x           // Parton momentum fraction
+      - std::array<double, 2> xf          // PDF value
+      - double scale [GeV]                // Factorisation scale
+    OneToManyRelations:
+      - edm4hep::MCParticle signal_vertex // pointing into MCParticle collection
```
tmadlener, line 811:
```suggestion
      - edm4hep::MCParticle signalVertex // pointing into MCParticle collection
```
Otherwise the getter will become `getSignal_vertex`

dirkzerwas:
There is one point you might want to talk to HepMC and the MC authors about: have a common approach to beam polarization:

- per particle: Spin Vector in EDM4HEP and theta, phi attributes in HepMC (if !=0) only is translatable to and fro
- Initial state beam polarization is not coded: Whizard does not set the spin vector for 100% polarization, in LCIO the information was written as parameter. It does not seem to be available in HepMC afaict.
- it would be great to have a common approach from HepMC and EDM4HEP together with MC authors (if they do not write it out, ......)

edm4hep.yaml
hegner: in the meantime that member moved. Did the equivalent rename.

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+
+  #---------- GeneratorEventParameters
+  edm4hep::GeneratorEventParameters:
+    Description: "Generator event parameters"
+    Author: "EDM4hep authors"
+    Members:
+      - double event_scale              // event scale
+      - double alphaQED                 // alpha_QED
+      - double alphaQCD                 // alpha_QCD
+      - int signal_process_id           // id of signal process
```
tmadlener, line 794:
```suggestion
      - int signalProcessId           // id of signal process
```
Just to conform to our usual naming scheme

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+      - double sqrt_s  [GeV]            // sqrt(s)
```
tmadlener, line 795:
```suggestion
      - double sqrtS  [GeV]            // sqrt(s)
```
Same here. Otherwise this will be `getSqrt_s`

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+      - double crossSectionErrors       // list of cross section errors
+    OneToManyRelations:
+      - edm4hep::MCParticle signalVertex // pointing into MCParticle collection
+
+  #---------- GeneratorPdfInfo
+  edm4hep::GeneratorPdfInfo:
+    Description: "Generator pdf information"
+    Author: "EDM4hep authors"
+    Members:
+      - std::array<int, 2> parton_id      // Parton PDG id
```
tmadlener, line 807:
```suggestion
      - std::array<int, 2> partonId      // Parton PDG id
```

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+      - std::array<int, 2> id             // LHAPDF PDF id (see https://lhapdf.hepforge.org/pdfsets.html)
```
tmadlener, line 808:
```suggestion
      - std::array<int, 2> lhapdfId             // LHAPDF PDF id (see https://lhapdf.hepforge.org/pdfsets.html)
```
(Or similar), to make it inferable from the name what you will get? Also if we keep it at `id` we will have an `id()` on the handle that will give you the `podio::ObjectId`, whereas the `getId()` method will get you this array.

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
 
+
+  #===== Generator related data =====
+
+  #---------- GeneratorEventParameters
+  edm4hep::GeneratorEventParameters:
+    Description: "Generator event parameters"
+    Author: "EDM4hep authors"
+    Members:
+      - double event_scale              // event scale
```
tmadlener, line 791:
```suggestion
      - double eventScale              // event scale
```

edm4hep.yaml
```diff
@@ -780,6 +780,36 @@ datatypes:
...
+      - double alphaQED                 // alpha_QED
+      - double alphaQCD                 // alpha_QCD
+      - int signal_process_id           // id of signal process
+      - double sqrt_s  [GeV]            // sqrt(s)
+    VectorMembers:
+      - double crossSections [pb]       // list of cross sections
+      - double crossSectionErrors       // list of cross section errors
```
tmadlener, line 798:
```suggestion
      - double crossSectionErrors [pb]      // list of cross section errors
```
I assume they are also given in `[pb]`, but in any case they should have a unit.

include/edm4hep/Constants.h
```diff
@@ -50,6 +50,14 @@ static constexpr const char* pidParameterNames = "ParameterNames";
 static constexpr const char* pidAlgoName = "AlgoName";
 static constexpr const char* pidAlgoType = "AlgoType";
 
+// Parameter names for Generator level metadata
+static constexpr const char* generatorToolVersionLabel = "generatorToolVersion";
+static constexpr const char* generatorToolNameLabel = "generatorToolName";
+static constexpr const char* generatorToolDescriptionLabel = "generatorToolDescription";
+static constexpr const char* generatorEventParametersLabel = "generatorEventParameters";
+static constexpr const char* generatorPdfInfoLabel = "generatorPdfInfo";
+static constexpr const char* generatorWeightNamesLabel = "generatorWeightNames";
```
tmadlener, line 59:
```suggestion
static constexpr const char* generatorToolVersion = "generatorToolVersion";
static constexpr const char* generatorToolName = "generatorToolName";
static constexpr const char* generatorToolDescription = "generatorToolDescription";
static constexpr const char* generatorEventParameters = "generatorEventParameters";
static constexpr const char* generatorPdfInfo = "generatorPdfInfo";
static constexpr const char* generatorWeightNames = "generatorWeightNames";
```
I would remove the `Label` bit from the name since we don't have that for the other constants either.

tmadlener (commented):
I have put in some consistency and style nitpicks.

include/edm4hep/Constants.h
hegner: I would rather add label to the other ones as well

tmadlener: At least some of them are already used for quite some time, so changing them would require a bit of work.

tmadlener (dismissed):

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,48 @@
...
+      name(name), version(version), description(description){};
+};
+
+const std::vector<GenToolInfo> getGenToolInfos(const podio::Frame& frame) {
+  auto toolInfos = std::vector<GenToolInfo>();
+  const auto names = frame.getParameter<std::vector<std::string>>(generatorToolNameLabel);
+  const auto versions = frame.getParameter<std::vector<std::string>>(generatorToolVersionLabel);
+  const auto descriptions = frame.getParameter<std::vector<std::string>>(generatorToolDescriptionLabel);
+  for (unsigned int i = 0; i < names.size(); ++i) {
+    toolInfos.emplace_back(names[i], versions[i], descriptions[i]);
```
tmadlener, line 27: This will need to take into account the changes that have happened in AIDASoft/podio#580 as `getParameter` now returns a `std::optional`

include/edm4hep/Constants.h
jmcarcell: I'm in favor of removing Label for consistency, that these are strings should something that a modern editor / IDE will tell you already. We also have right now all the constants except one start with upper case both the variable name and the string so maybe we want to keep some sort of style here.

tmadlener: I am also in favor of both, i.e. upper case variable names and values and removing the `Label` bit.

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,50 @@
...
+      name(name), version(version), description(description){};
+};
+
+namespace utils {
+
+  const std::vector<GenToolInfo> getGenToolInfos(const podio::Frame& frame) {
+    auto toolInfos = std::vector<GenToolInfo>();
+    const auto names = frame.getParameter<std::vector<std::string>>(generatorToolName).value();
+    const auto versions = frame.getParameter<std::vector<std::string>>(generatorToolVersion).value();
+    const auto descriptions = frame.getParameter<std::vector<std::string>>(generatorToolDescription).value();
```
tmadlener, line 27: In case any of these parameters is not available this implementation will throw a `std::bad_optional_access`, which might be a bit cryptic, but would at least be in line with "fail early in case of problems".

If we wanted a slightly more relaxed approach without exceptions escaping, we could do:
```suggestion
    const auto names = frame.getParameter<std::vector<std::string>>(generatorToolName).value_or({});
    const auto versions = frame.getParameter<std::vector<std::string>>(generatorToolVersion).value_or({});
    const auto descriptions = frame.getParameter<std::vector<std::string>>(generatorToolDescription).value_or({});
```
In that case the worst that can happen is an empty vector, which probably also shows that something went wrong.

I am in favor of the current implementation as that makes errors harder to ignore and easier to diagnose (even if the `std::bad_optional_access` might be a bit cryptic).

include/edm4hep/Constants.h
```diff
@@ -50,6 +50,14 @@ static constexpr const char* pidParameterNames = "ParameterNames";
 static constexpr const char* pidAlgoName = "AlgoName";
 static constexpr const char* pidAlgoType = "AlgoType";
 
+// Parameter names for Generator level metadata
+static constexpr const char* generatorToolVersion = "generatorToolVersion";
+static constexpr const char* generatorToolName = "generatorToolName";
+static constexpr const char* generatorToolDescription = "generatorToolDescription";
+static constexpr const char* generatorEventParameters = "generatorEventParameters";
+static constexpr const char* generatorPdfInfo = "generatorPdfInfo";
+static constexpr const char* generatorWeightNames = "generatorWeightNames";
```
tmadlener, line 59:
```suggestion
static constexpr const char* GeneratorToolVersion = "GeneratorToolVersion";
static constexpr const char* GeneratorToolName = "GeneratorToolName";
static constexpr const char* GeneratorToolDescription = "GeneratorToolDescription";
static constexpr const char* GeneratorEventParameters = "GeneratorEventParameters";
static constexpr const char* GeneratorPdfInfo = "GeneratorPdfInfo";
static constexpr const char* GeneratorWeightNames = "GeneratorWeightNames";
```
As suggested by @jmcarcell for more internal consistency.

test/read_events.h
```diff
@@ -16,6 +19,27 @@
...
+  if (toolinfo.name != "something")
+    throw std::runtime_error("toolinfo.name != 'something'");
+  if (toolinfo.version != "v1")
+    throw std::runtime_error("toolinfo.version != 'v1'");
+  if (toolinfo.description != "some tool")
+    throw std::runtime_error("toolinfo.description != 'some tool'");
+
+  //===============================================================================
+  // get generator weight names
+  auto weightNames = run.getParameter<std::vector<std::string>>(edm4hep::generatorWeightNames);
```
tmadlener, line 36:
```suggestion
  auto weightNames = run.getParameter<std::vector<std::string>>(edm4hep::generatorWeightNames).value();
```

Looks like I missed this usage of parameters in the last review.

include/edm4hep/GenToolInfo.h
tmadlener: After some more discussion with @dirkzerwas about how this will be used, I think we can do the "error" handling internally, and simply return an empty vector instead of letting the exception propagate. The main reasons are
- `vector::empty` still let's users check whether information was available
- Information on the tools rather falls into a "nice to have" category, rather than crucially important information for event processing.

hegner: `value_or` should do then the right thing then

include/edm4hep/Constants.h
```diff
@@ -50,6 +50,14 @@ static constexpr const char* pidParameterNames = "ParameterNames";
 static constexpr const char* pidAlgoName = "AlgoName";
 static constexpr const char* pidAlgoType = "AlgoType";
 
+// Parameter names for Generator level metadata
+static constexpr const char* GeneratorToolVersion = "GeneratorToolVersion";
+static constexpr const char* GeneratorToolName = "GeneratorToolName";
+static constexpr const char* GeneratorToolDescription = "GeneratorToolDescription";
+static constexpr const char* GeneratorEventParameters = "GeneratorEventParameters";
+static constexpr const char* GeneratorPdfInfo = "GeneratorPdfInfo";
```
tmadlener, line 58: These are now clashing with the generated classes. Maybe all of these should just be prefixed with `Gen` instead of `Generator`?

include/edm4hep/GenToolInfo.h
```diff
@@ -0,0 +1,50 @@
...
+#define EDM4HEP_GENTOOLINFO_H
+
+#include "edm4hep/Constants.h"
+#include "podio/Frame.h"
+#include <string>
+#include <vector>
+
+namespace edm4hep {
+
+struct GenToolInfo {
```
tmadlener, line 11: The classes we define in the yaml file are all prefixed `Generator`, but this one is `Gen`. For consistency reasons, I think they should both be the same. The choice is also related with how we want to name the constants.

tmadlener (commented):
A few more things triggered by the clashing variable and class names.

include/edm4hep/Constants.h
hegner: I think I will do that indeed.

andresailer: Or we add a namespace for the constants.
And we copy the existing constants into the namespace and deprecate the namespaceless ones (we can even align spelling when we do that without breaking anything)

hegner: that would be a major change and a bit cumbersome if dealing e.g. with the matrices end the corresponding enums. Do you really want to do that?

m-fila: IIRC the enums were supposed to stay where they are, since they aren't constants

tmadlener: I would like a `edm4hep::label` namespace for things that are a label. the enums for covariance matrix access can / should stay in the `edm4hep` namespace.

tmadlener: See #315 for a proposal with a new `labels` namespace.

tmadlener:
Resolved the merge conflicts, added a bit of documentation and renamed `GenToolInfo` -> `GeneratorToolInfo` for consistency.

hegner:
Perfect. Thank you

tmadlener:
Unless there are complaints by tomorrows meeting, I will merge this shortly before that.

dirkzerwas:
@tmadlener 
include/edm4hep/GenToolInfo.h
you might need to swiitch to plural for:
frame.putParameter(GeneratorToolNames, std::move(names));
frame.putParameter(GeneratorToolVersions, std::move(versions));
frame.putParameter(GeneratorToolDescriptions, std::move(descriptions));
and for the triplet of frame.getParameter(....)

include/edm4hep/GeneratorToolInfo.h
```diff
@@ -0,0 +1,88 @@
...
+  ///          found an empty vector will be returned.
+  const inline std::vector<GeneratorToolInfo> getGenToolInfos(const podio::Frame& frame) {
+    using namespace edm4hep::labels;
+    auto toolInfos = std::vector<GeneratorToolInfo>();
+    const auto names =
+        frame.getParameter<std::vector<std::string>>(GeneratorToolName).value_or(std::vector<std::string>{});
+    const auto versions =
+        frame.getParameter<std::vector<std::string>>(GeneratorToolVersion).value_or(std::vector<std::string>{});
+    const auto descriptions =
+        frame.getParameter<std::vector<std::string>>(GeneratorToolDescription).value_or(std::vector<std::string>{});
```
tmadlener, line 56:
```suggestion
        frame.getParameter<std::vector<std::string>>(GeneratorToolNames).value_or(std::vector<std::string>{});
    const auto versions =
        frame.getParameter<std::vector<std::string>>(GeneratorToolVersions).value_or(std::vector<std::string>{});
    const auto descriptions =
        frame.getParameter<std::vector<std::string>>(GeneratorToolDescriptions).value_or(std::vector<std::string>{});
```

include/edm4hep/GeneratorToolInfo.h
```diff
@@ -0,0 +1,88 @@
...
+    for (auto& toolInfo : toolInfos) {
+      names.push_back(toolInfo.name);
+      versions.push_back(toolInfo.version);
+      descriptions.push_back(toolInfo.description);
+    }
+
+    using namespace edm4hep::labels;
+    frame.putParameter(GeneratorToolName, std::move(names));
+    frame.putParameter(GeneratorToolVersion, std::move(versions));
+    frame.putParameter(GeneratorToolDescription, std::move(descriptions));
```
tmadlener, line 83:
```suggestion
    frame.putParameter(GeneratorToolNames, std::move(names));
    frame.putParameter(GeneratorToolVersions, std::move(versions));
    frame.putParameter(GeneratorToolDescriptions, std::move(descriptions));
```

PR #311, opened 2024-05-28
Remove the dQ/dx information from the track

tmadlener:
BEGINRELEASENOTES
- Remove the `dQ/dx` information from the Track
- Add `TrackPIDHandler` to allow for the reverse lookup

ENDRELEASENOTES

- [x] Needs check of things that break downstream
	- [x] https://github.com/key4hep/k4FWCore/pull/198
- [x] Currently breaks in the following packages
  - k4EDM4hep2LcioConv (https://github.com/key4hep/k4EDM4hep2LcioConv/pull/69)
  - k4SimDelphes (https://github.com/key4hep/k4SimDelphes/pull/121)
  - FCCAnalyses (https://github.com/HEP-FCC/FCCAnalyses/pull/387)
  - k4RecTracker (https://github.com/key4hep/k4RecTracker/pull/23)


tmadlener:
Make a dedicated type for the dE/dx (or dN/dx) information and don't reuse the `RecDqdx`.

gaede:
I think we do not want the dEdx or DQdx in the Track anymore but rather a type (e.g. TrackDQdx) that points back to the Track. Otherwise we have the same issue as w/ dEdx and having to copy the Track when computing this...

tmadlener:
The FCCAnalyses part will most likely be rather involved, as some of the logic that is used there will simply no longer work as easily.

tmadlener:
As discussed in todays meeting: prioritize an EDM4hep pre-release and for the moment turn-off the dE/dx features in FCCAnalyses and bring them back to live after the pre-release.

tmadlener:
With all the PRs mentioned in the top level comment, I am able to build the complete stack locally. I will merge all of them in the afternoon, to get them into the nightlies for tomorrow.

PR #315, opened 2024-06-13
Introduce a `labels` namespace for the string constants

tmadlener:
BEGINRELEASENOTES
- Introduce the `edm4hep::labels` namespace to hold the string constants that are used for consistent labeling / naming of collections or parameters.
  - Use this opportunity to homogenize the naming and capitalization of the variables but leave string constants unchanged
- Deprecate existing string constants

ENDRELEASENOTES

As triggered by [this discussion](https://github.com/key4hep/EDM4hep/pull/310#discussion_r1636352789)

### DIFF
--- a/include/edm4hep/Constants.h
+++ b/include/edm4hep/Constants.h
@@ -41,7 +41,7 @@ namespace labels {
 
 DEPRECATED_LABEL(CellIDEncoding, CellIDEncoding);
 DEPRECATED_LABEL(EventHeaderName, EventHeader);
-DEPRECATED_LABEL(EventWeights, EventWeights);
+DEPRECATED_LABEL(EventWeights, EventWeightsNames);
 DEPRECATED_LABEL(shapeParameterNames, ShapeParameterNames);
 DEPRECATED_LABEL(EventFilterStats, EventFilterStats);
 
--- a/include/edm4hep/Constants.h
+++ b/include/edm4hep/Constants.h
@@ -21,12 +21,33 @@
 
 #include <cstdint>
 
+#define DEPRECATED_LABEL(name, newname)                                                                                \
+  static constexpr const auto name [[deprecated("Use 'edm4hep::labels::" #newname "' instead")]] = labels::newname
+
 namespace edm4hep {
-static constexpr const char* CellIDEncoding = "CellIDEncoding";
-static constexpr const char* EventHeaderName = "EventHeader";
-static constexpr const char* EventWeights = "EventWeightNames";
-static constexpr const char* shapeParameterNames = "shapeParameterNames";
-static constexpr const char* EventFilterStats = "EventFilterStats";
+namespace labels {
+  static constexpr const char* CellIDEncoding = "CellIDEncoding";
+  static constexpr const char* EventHeader = "EventHeader";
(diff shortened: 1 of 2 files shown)


include/edm4hep/Constants.h
```diff
@@ -21,12 +21,33 @@
...
 namespace edm4hep {
-static constexpr const char* CellIDEncoding = "CellIDEncoding";
-static constexpr const char* EventHeaderName = "EventHeader";
-static constexpr const char* EventWeights = "EventWeightNames";
-static constexpr const char* shapeParameterNames = "shapeParameterNames";
-static constexpr const char* EventFilterStats = "EventFilterStats";
+namespace labels {
+  static constexpr const char* CellIDEncoding = "CellIDEncoding";
+  static constexpr const char* EventHeader = "EventHeader";
+  static constexpr const char* EventWeights = "EventWeightNames";
```
andresailer, line 31:
```suggestion
  static constexpr const char* EventWeightsNames = "EventWeightNames";
```
I would align that with the others

jmcarcell: I don't think we need to include `Names` now that is namespaced and that information is already encoded in the namespace. Before they were in the global namespace and that information was only in `Names`

include/edm4hep/Constants.h
```diff
@@ -21,12 +21,33 @@
...
+#define DEPRECATED_LABEL(name, newname)                                                                                \
+  static constexpr const auto name [[deprecated("Use 'edm4hep::labels::" #newname "' instead")]] = labels::newname
+
 namespace edm4hep {
-static constexpr const char* CellIDEncoding = "CellIDEncoding";
-static constexpr const char* EventHeaderName = "EventHeader";
-static constexpr const char* EventWeights = "EventWeightNames";
-static constexpr const char* shapeParameterNames = "shapeParameterNames";
-static constexpr const char* EventFilterStats = "EventFilterStats";
+namespace labels {
```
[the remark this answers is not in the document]
hegner: the plural makes more sense here as it contains multiple items. With the `utils` namespace we do the same. So we should keep it that way.

hegner (approved):
Fine from my side. Given we can check this fix once the unrelated problem introduced by #308 is fixed

include/edm4hep/Constants.h
```diff
@@ -21,12 +21,33 @@
...
+  /// The collection parameter name for accessing the names of the parameters for
+  /// a ParticleID collection
+  static constexpr const char* PIDParameterNames = "ParameterNames";
+  static constexpr const char* PIDAlgoName = "AlgoName";
+  static constexpr const char* PIDAlgoType = "AlgoType";
+} // namespace labels
+
+DEPRECATED_LABEL(CellIDEncoding, CellIDEncoding);
+DEPRECATED_LABEL(EventHeaderName, EventHeader);
+DEPRECATED_LABEL(EventWeights, EventWeights);
```
tmadlener, line 44:
```suggestion
DEPRECATED_LABEL(EventWeights, EventWeightsNames);
```
Necessary fix after renaming the new label.

andresailer (approved):
Looks good to me. Thanks!

Issue #319, opened 2024-06-18
`radiusOfInnerMostHit` could be removed from `Track`


tmadlener:
The `Track` currently has a member `radiusOfInnerMostHit`, which could be removed, as it is technically easily computable from the attached `TrackerHit`s. The only / main(?) reason to keep it would be to keep this information around even if the `TrackerHit`s are dropped (as is the case for example in DSTs for ILD).

gaede:
We (BH, JMC, FG) believe that this should be droped.  (If really needed later, adding it is easy w/ schema evolution).

Zehvogel:
This was quite useful information on DST files (no-hits) imo...

tmadlener:
Removing it for now, since it's mixing analysis and reconstruction level quantities too much. We have a similar discussion for other items, e.g. #322

andresailer:
Dear @farakiko 

Please open a new issue, and please don't screenshoot text, but give us the actual code as text so we don't have to potentially type things ourselves.

Thanks,
Andre

farakiko:
Thank you! Submitted the issue [here](https://github.com/key4hep/EDM4hep/issues/384).

PR #326, opened 2024-06-20
Remove `radiusOfInnermostHit` from tracks

jmcarcell:
as discussed in https://github.com/key4hep/EDM4hep/issues/319. 

Changes in other repos:
- https://github.com/key4hep/k4FWCore/pull/202
- https://github.com/key4hep/k4SimDelphes/pull/119
- https://github.com/key4hep/k4EDM4hep2LcioConv/pull/72

BEGINRELEASENOTES
- Remove `radiusOfInnermostHit` from tracks

ENDRELEASENOTES


tmadlener:
The fact that this requires changes in `k4SimDelphes` makes me think this will also show up in some analyses in FCCAnalyses, but I haven't checked.

jmcarcell:
I checked and didn't find anything in FCCAnalyses

tmadlener:
OK. Then maybe it was there only temporarily, or we just filled it in k4SimDelphes because we could. Thanks for checking.

hegner (approved):

hegner:
Perfect. Good we checked this again :)

hegner:
can we maybe finally fix the release build test environment? Otherwise I am fine with merging

tmadlener:
Fixing the release build would require a new Key4hep release. For EDM4hep we could in principle pick up the latest podio tag, but for other packages, we would also need a new EDM4hep tag. I am not sure if we want to make such a tag at the moment.

PR #329, opened 2024-06-26
Make the Vertex have a proper `type` bitfield instead of a `primary` field

### LINKED ISSUE #328: Make the `Vertex::primary` field a proper bit field
(this issue is mainly intended to keep track of a discussion that already happened)

As brought up by @kjvbrt, the current `primary` field in the `Vertex` class is documented to be a boolean field, but is actually implemented via an int:

https://github.com/key4hep/EDM4hep/blob/4f72168fb5ec1abe49f8ff67d45d766c88048795/edm4hep.yaml#L551-L555

We decided today to make this member a proper bitfield, renaming it to `type` (in accordance with similar fields in other datatypes) and add a few reserved bits and `ExtraCode` similar to what is currently present in `MCParticle`. The reserved bit values that we add for now are `primary`, `secondary` and `tertiary`. More reserved bits can be added in the future without affecting the schema.

tmadlener:
BEGINRELEASENOTES
- Rename the `Vertex` `primary` field to `type` and make it a 32 bit unsigned integer
- Introduce 3 reserved bits and accompanying utility / ExtraCode functionality for checking primary, secondary or tertiary vertex
- Introduce `setBit` and `checkBit` utility functionality and also use that for the `MCParticle` extra code

ENDRELEASENOTES

Fixes #328 

- [ ] Needs check / changes for downstream consumers (these have to go in simultaneously)
  - https://github.com/key4hep/k4EDM4hep2LcioConv/pull/74 (ready for merge)
  - https://github.com/HEP-FCC/FCCAnalyses/pull/382 (under review)

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -537,7 +537,7 @@ datatypes:
     Description: "Vertex"
     Author: "EDM4hep authors"
     Members:
-      - int32_t                 primary       // boolean flag, if vertex is the primary vertex of the event
+      - uint32_t            type          // Flagword that defines the type of the vertex, see reserved bits for more information
       - float               chi2          // chi-squared of the vertex fit
       - int32_t             ndf           // number of degrees of freedom of the vertex fit
       - edm4hep::Vector3f   position      // [mm] position of the vertex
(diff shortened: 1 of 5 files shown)


edm4hep.yaml
```diff
@@ -563,16 +563,34 @@ datatypes:
...
-      includes: "#include <edm4hep/Constants.h>"
+      includes: "#include <edm4hep/Constants.h>\n
+                 #include <edm4hep/utils/bit_utils.h>\n"
       declaration: "
       /// Get the position covariance matrix value for the two passed dimensions\n
       float getCovMatrix(edm4hep::Cartesian dimI, edm4hep::Cartesian dimJ) const { return getCovMatrix().getValue(dimI, dimJ); }\n
+      // Reserved bits for the type flagword\n
+      static constexpr int BITPrimaryVertex = 1;\n
+      static constexpr int BITSecondaryVertex = 2;\n
+      static constexpr int BITTertiaryVertex = 2;\n
```
Zehvogel, line 574:
```suggestion
      static constexpr int BITTertiaryVertex = 3;\n
```
I guess this is not supposed to be the same as `BITSecondaryVertex`?

Zehvogel:
LCFIPlus already distinguishes between secondary (tertiary etc.) vertices and V0 vertices (although by writing them into different collections). We could add V0 another bit for V0 directly imo.

edm4hep.yaml
tmadlener: Copy-paste :see_no_evil: Thanks.

tmadlener:
Does LCFIPlus only split the collections between V0 and non-V0, or also between primary, secondary (, tertiary)?

Zehvogel:
3 collections: primary, buildup, buildup_V0

PR #330, opened 2024-06-26
Remove no longer relevant bit of docstring relating to `Track::type`

tmadlener:
BEGINRELEASENOTES
- Remove docstring of `Track::type` that is no longer relevant for EDM4hep and was a leftover from LCIO.
  - EDM4hep does not (yet) reserve any bits in the Track

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -533,7 +533,7 @@ datatypes:
     Description: "Reconstructed track"
     Author: "EDM4hep authors"
     Members:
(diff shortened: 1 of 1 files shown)


jmcarcell:
Does this also apply to Clusters?

tmadlener:
Good catch. Yes it does. Fixed

jmcarcell (approved):

Issue #323, opened 2024-06-18
Better integration of datatypes for the drift chamber study of CEPC into EDM4hep


tmadlener:
There are a few inconsistencies in the datatypes that were introduced in https://github.com/key4hep/EDM4hep/pull/179 for the drift chamber study. A tangential but related discussion is also in #322. Opening this issue as a sort of summary discussion board to see if this situation can be improved, also keeping in mind the proposed new datatypes in #299 

My main complaints for these are:
- `HitLevelData` is a rather generic lwo level datatype to store pretty much any detector hit / readout. However, it has a few members that make it very gaseous detector specific. I think this should be reflected in the name
https://github.com/key4hep/EDM4hep/blob/bd9f45039149781a9c54799a6d03df56b1b5ed49/edm4hep.yaml#L239-L244

- `Hypothesis` is not properly documented and seems to be lacking a `ndf` field(?). Specifically it has a `chi2` value, but it's not really documented what this `chi2` value represents. Also the `sigma` docstring could be refined.
https://github.com/key4hep/EDM4hep/blob/bd9f45039149781a9c54799a6d03df56b1b5ed49/edm4hep.yaml#L232-L236

- Overall the structure of the dataypes is fairly involved, but also quite separate from the rest of EDM4hep and I am wondering whether there is a way to (re-)organize the information in a cleaner way in order to make them better integrated into the rest of EDM4hep. Relation diagram of the datatypes and some EDM4hep datatypes:
![image](https://github.com/key4hep/EDM4hep/assets/16774861/b8c4a6b5-1876-4acf-a92a-2b960bafccb0)

tmadlener:
This is a brief summary the outcome of todays meeting, where we also discussed this (@mirguest, please correct me if anything is wrong in the following).

CEPCSW is currently using a "frozen" version of EDM4hep and other externals since they are heavily involved in preparing the TDR. As such, at least from that point of view we (as in EDM4hep) are pretty much at liberty to break things as far as CEPCSW is concerned. We have then decided to remove some of the drift chamber study related datatypes for EDM4hep v1.0 and bring them back in a more consistent form afterwards. For that we also need to consult with IDEA / FCC reconstruction developers (@BrieucF just as a heads up) and check how to bring everything into a consistent shape, also considering the discussion / resolution in #322. CEPCSW should then be able to "catch up" / migrate from the TDR status to EDM4hep v1.0, once the TDR is done.

Keeping this open, but moving it out of the EDM4hep v1.0 project.

Issue #322, opened 2024-06-18
General discussion on design philosophy for mixing reconstruction and analysis


tmadlener:
There are a few places in EDM4hep (including a few places where similar things are proposed, as in #299), where there is a bit of mixing between analysis / interpretation of quantities and "pure" reconstruction. This issue should serve as a discussion board to figure out if we are OK with this, or if we want to / can come up with a different approach.

Using `RecDqdx` here as a specific example, but a similar thing also appears in the proposed TOF datatypes (#299).

https://github.com/key4hep/EDM4hep/blob/bd9f45039149781a9c54799a6d03df56b1b5ed49/edm4hep.yaml#L773-L780

That stores an array of 5 hypothesis for different particle types. This is rather strongly mixing the reconstruction of a value, in this case `dE/dx` or `dN/dx` with its later interpretation.

There is nothing strictly speaking against this, but IMHO it mixes reconstruction and analysis rather heavily and it might be better to split these apart slightly.

PR #333, opened 2024-06-27
Remove some of the drift chamber study types again

tmadlener:
BEGINRELEASENOTES
- Remove some of the drift chamber study types that were introduced in [#179](https://github.com/key4hep/EDM4hep/pull/179) again since they need a bit more consideration
  - The removed types and components are: `Hypothesis`, `HitLevelData`, `RecIonizationCluster`, `SimPrimaryIonizationCluster`, `TrackerPulse`
  - Keep the types and components that are already in use, mainly `TimeSeries`
- Remove the usage of the removed components from `RecDqdx` to make it a purely reconstruction level type
- Fix script that checks that all types are part of the json dumper

ENDRELEASENOTES

See #322, #323 for discussion on how we arrived at this point.

The changes for `RecDqdx` are also present in #311, leading to a likely merge conflict depending on which PR is merged first

We keep the `TimeSeries` as it is used already, e.g. in dual-readout:
- https://github.com/HEP-FCC/dual-readout/blob/51f9766bb3f9e80644bb032b98f1f5b1d4cb0327/DRdigi/include/DigiSiPM.h#L31

- [x] Downstream checks & fixes
  - https://github.com/key4hep/k4FWCore/pull/205

### DIFF
--- a/test/tools/test_all_collections.py
+++ b/test/tools/test_all_collections.py
@@ -21,8 +21,7 @@ def test(yamlfile_path, cxxfile_path):
     datatypes_found = []
 
     with open(cxxfile_path, mode='r', encoding="utf-8") as cxxfile:
-        for cxxline in cxxfile:
-            cxxline = cxxfile.readline()
+        for cxxline in cxxfile.readlines():
             result = re.search('insertIntoJson<edm4hep::(.+?)Collection>',
                                cxxline)
             if result:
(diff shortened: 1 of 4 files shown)


test/tools/test_all_collections.py
```diff
@@ -21,8 +21,7 @@ def test(yamlfile_path, cxxfile_path):
     datatypes_found = []
 
     with open(cxxfile_path, mode='r', encoding="utf-8") as cxxfile:
-        for cxxline in cxxfile:
-            cxxline = cxxfile.readline()
+        for cxxline in cxxfile.readlines():
```
tmadlener, line 24: This prevents accidentally skipping every other line(!), which worked before because we were and now we removed an odd number of lines, so it broke for this PR.

PR #334, opened 2024-06-28
Remove compatibility with podio versions before v1.0

### LINKED ISSUE #306: Remove compatibility with podio < 1.0
This is a summary issue to keep track of the places where we have some special casing for pre 1.0 podio versions. These can obviously only be addressed one podio v1.0 is out.

- Pre processor checks introduced in #305 
- Pre processor checks to switch from `ROOTFrame{Reader,Writer}` to `ROOT{Reader,Writer}` introduced in #263 and #293

tmadlener:
BEGINRELEASENOTES
- Remove compatibility with podio versions before 1.0

ENDRELEASENOTES

Since we already bumped this in our CMake config (https://github.com/key4hep/EDM4hep/pull/327), we can also remove this from our code.

Fixes #306

### DIFF
--- a/test/hepmc/edm4hep_testhepmc.cc
+++ b/test/hepmc/edm4hep_testhepmc.cc
@@ -17,11 +17,7 @@
 #include "HepPDT/ParticleID.hh"
 
 #include "podio/Frame.h"
-#if PODIO_VERSION_MAJOR > 0 || (PODIO_VERSION_MAJOR == 0 && PODIO_VERSION_MINOR >= 99)
 #include "podio/ROOTWriter.h"
-#else
-#include "podio/ROOTFrameWriter.h"
-#endif
 
 #include "edm4hep/MCParticleCollection.h"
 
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -235,8 +235,7 @@ spack-*
 install/
 ## podio
 podio_generated_files.cmake
-/include/podio/podioVersion.h
-/python/podio/__init__.py
```
jmcarcell, line 239: These come from https://github.com/key4hep/key4hep-dev-utils/blob/main/defaults/.gitignore#L238, I think copied and pasted from the `.gitignore` in podio. The up to date ones seem to be 
```
/python/podio/__version__.py
/python/podio_version.py
```

jmcarcell:
I opened other PRs with the same changes in other repositories:
- https://github.com/key4hep/k4FWCore/pull/206
- https://github.com/key4hep/k4SimDelphes/pull/120
- https://github.com/key4hep/k4EDM4hep2LcioConv/pull/80

I think it's fine to merge them but they can also wait. The only case when keeping this code is useful is when someone wants to build the current versions on top of older versions of EDM4hep and Podio but not so old they don't work.

.gitignore
tmadlener: Ah, this is the global .gitignore, right? Why are we having podio specific ignores there? These should not appear in any repository, and the podio .gitignore is not globally managed.

jmcarcell: Ok they should be removed now, the changes here can be kept or not, they will be overwritten at some point anyway

jmcarcell (approved):

Issue #338, opened 2024-07-10
Python bindings for mac not working due to different library suffix


tmadlener:
The python bindings are not working on macOS, because the suffix for shared libraries is `dylib` instead of `.so`. We currently explicitly go looking for `.so` though in our `__init__.py`:

https://github.com/key4hep/EDM4hep/blob/3d271d0ae614870d2118e5c078207642fb2d266b/python/edm4hep/__init__.py#L6-L8

It might be enough to simply remove the suffix and let ROOT do the OS specific handling, but I am not sure.

PR #339, opened 2024-07-10
Load libraries without the .so suffix to make it work in other OSes

jmcarcell:
Should fix https://github.com/key4hep/EDM4hep/issues/338

BEGINRELEASENOTES
- Load libraries without the .so suffix to make it work in other OSes

ENDRELEASENOTES

### DIFF
--- a/python/edm4hep/__init__.py
+++ b/python/edm4hep/__init__.py
@@ -3,13 +3,13 @@ import sys
 
 from .__version__ import __version__
 import ROOT
-res = ROOT.gSystem.Load('libedm4hepDict.so')
+res = ROOT.gSystem.Load('libedm4hepDict')
 if res < 0:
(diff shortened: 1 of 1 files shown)


tmadlener:
I got a report that things work on mac OS with this.

jmcarcell:
Ok I was going to test it but then I don't; it looks like this is how loading libraries is done in ROOT most of the time unless they are Linux-specific, and locally it works for me on Linux.

tmadlener:
I also seem to remember documentation about this being the recommended way to do this for platform agnostic support, but I can't find that right now.

tmadlener (approved):

jmcarcell:
There is this: https://github.com/root-project/root/blob/master/core/unix/src/TUnixSystem.cxx#L4803 that is telling us which suffixes are being used

PR #342, opened 2024-07-21
Drop support for the ROOTLegacyReader when converting to JSON

jmcarcell:
since legacy files were done with incompatible versions of EDM4hep.

BEGINRELEASENOTES
- Drop support for the ROOTLegacyReader when converting to JSON

ENDRELEASENOTES

### DIFF
--- a/tools/include/edm4hep2json.hxx
+++ b/tools/include/edm4hep2json.hxx
@@ -191,7 +191,7 @@ int read_frames(const std::string& filename, const std::string& jsonFile, const 
 
   auto collList = splitString(requestedCollections);
   if (collList.empty()) {
-    auto frame = podio::Frame(reader.readFrame(frameName, 0));
+    auto frame = reader.readFrame(frameName, 0);
     collList = frame.getAvailableCollections();
   }
   if (collList.empty()) {
@@ -267,13 +267,13 @@ int read_frames(const std::string& filename, const std::string& jsonFile, const 
         std::cout << "INFO: Reading event " << i << std::endl;
       }
 
-      auto frame = podio::Frame(reader.readFrame(frameName, i));
+      auto frame = reader.readFrame(frameName, i);
       auto eventDict = processEvent(frame, collList, reader.currentFileVersion());
       allEventsDict["Event " + std::to_string(i)] = eventDict;
     }
   } else {
     for (auto& i : eventVec) {
-      auto frame = podio::Frame(reader.readFrame(frameName, i));
+      auto frame = reader.readFrame(frameName, i);
       auto eventDict = processEvent(frame, collList, reader.currentFileVersion());
       allEventsDict["Event " + std::to_string(i)] = eventDict;
     }
--- a/tools/include/edm4hep2json.hxx
+++ b/tools/include/edm4hep2json.hxx
@@ -34,6 +34,7 @@
 
 // podio specific includes
 #include "podio/Frame.h"
+#include "podio/Reader.h"
 #include "podio/UserDataCollection.h"
 #include "podio/podioVersion.h"
 
@@ -170,13 +171,10 @@ std::vector<std::string> splitString(const std::string& inString) {
 
   return outString;
 }
-
-template <typename ReaderT>
 int read_frames(const std::string& filename, const std::string& jsonFile, const std::string& requestedCollections,
                 const std::string& requestedEvents, const std::string& frameName, int nEventsMax = -1,
                 bool verboser = false) {
-  ReaderT reader;
-  reader.openFile(filename);
+  podio::Reader reader = podio::makeReader(filename);
 
   nlohmann::json allEventsDict;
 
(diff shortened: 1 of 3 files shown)


tools/src/edm4hep2json.cxx
```diff
@@ -120,20 +119,8 @@ int main(int argc, char** argv) {
...
-  if (legacyReader) {
-    std::cout << "WARNING: Reading legacy file, some collections might not be recognized!" << std::endl;
-    return read_frames<podio::ROOTLegacyReader>(inFilePath, outFilePath, requestedCollections, requestedEvents,
-                                                frameName, nEventsMax, verboser);
-  } else {
-    return read_frames<podio::ROOTReader>(inFilePath, outFilePath, requestedCollections, requestedEvents, frameName,
-                                          nEventsMax, verboser);
-  }
+  return read_frames<podio::ROOTReader>(inFilePath, outFilePath, requestedCollections, requestedEvents, frameName,
+                                        nEventsMax, verboser);
```
tmadlener, line 123: Maybe switch this directly to the `podio::Reader`?

jmcarcell: To do this one option to change the category of the dumped frame has to be changed, but maybe that option is never used. Would that be fine @kjvbrt?

tmadlener: Why doe we need to change that? The `Reader` also has a `readFrame` method [doc](https://key4hep.web.cern.ch/podio/cpp_api/class/classpodio_1_1_reader.html#_CPPv4N5podio6Reader9readFrameERKNSt6stringE6size_t). We could get rid of a few double Frame constructions, since the `Reader` already returns a Frame.

jmcarcell: Yes, fixed now

tools/include/edm4hep2json.hxx
```diff
@@ -196,7 +193,7 @@ int read_frames(const std::string& filename, const std::string& jsonFile, const
 
   auto collList = splitString(requestedCollections);
   if (collList.empty()) {
-    auto frame = podio::Frame(reader.readEntry(frameName, 0));
+    auto frame = podio::Frame(reader.readFrame(frameName, 0));
```
tmadlener, line 196:
```suggestion
    auto frame = reader.readFrame(frameName, 0);
```

tools/include/edm4hep2json.hxx
```diff
@@ -272,13 +269,13 @@ int read_frames(const std::string& filename, const std::string& jsonFile, const
         std::cout << "INFO: Reading event " << i << std::endl;
       }
 
-      auto frame = podio::Frame(reader.readEntry(frameName, i));
+      auto frame = podio::Frame(reader.readFrame(frameName, i));
```
tmadlener, line 272:
```suggestion
      auto frame = reader.readFrame(frameName, i);
```

tools/include/edm4hep2json.hxx
```diff
@@ -272,13 +269,13 @@ int read_frames(const std::string& filename, const std::string& jsonFile, const
...
       auto eventDict = processEvent(frame, collList, reader.currentFileVersion());
       allEventsDict["Event " + std::to_string(i)] = eventDict;
     }
   } else {
     for (auto& i : eventVec) {
-      auto frame = podio::Frame(reader.readEntry(frameName, i));
+      auto frame = podio::Frame(reader.readFrame(frameName, i));
```
tmadlener, line 278:
```suggestion
      auto frame = reader.readFrame(frameName, i);
```

tools/src/edm4hep2json.cxx
```diff
@@ -120,20 +115,7 @@ int main(int argc, char** argv) {
...
   }
 
-  bool legacyReader = false;
-  {
-    std::unique_ptr<TFile> inFile(TFile::Open(inFilePath.c_str(), "READ"));
-    legacyReader = !inFile->GetListOfKeys()->FindObject("podio_metadata");
-  }
-
-  if (legacyReader) {
-    std::cout << "WARNING: Reading legacy file, some collections might not be recognized!" << std::endl;
```
tmadlener, line 130: Can we keep the check and a slightly adapted warning message that tells people that they cannot use this file because it has not been done with a compatible EDM4hep version?

tools/include/edm4hep2json.hxx
jmcarcell: Fixed

tools/include/edm4hep2json.hxx
jmcarcell: Fixed

tools/include/edm4hep2json.hxx
jmcarcell: Fixed

tools/src/edm4hep2json.cxx
jmcarcell: Check added back and modified, now it's an error since it's likely going to crash anyway.

PR #346, opened 2024-07-28
Update LCG workflow to use LCG_106 instead of LCG_104

jmcarcell:
BEGINRELEASENOTES
- Update LCG workflow to use LCG_106 instead of LCG_104

ENDRELEASENOTES


tmadlener (approved):

tmadlener:
I suppose part of the plan in this case is also to get rid of some of the issues in #272? I would still adjust the `PYTHONPATH` after installing podio in this workflow, because otherwise we might run into similar issues in the future, if we make changes / additions to the podio python interface.

jmcarcell:
Yeah, anyway it's slightly old 104 so we can just use 106

PR #347, opened 2024-07-29
Check for the size of the input for one of the constructors of the covariance matrices

jmcarcell:
currently it's possible to set only the first N elements of the matrix. I think typically this will be mostly a mistake rather than intentionally filling only the first N.

Also static_cast to float, because with older GCC and Clang this is fine, newer GCC throws a warning and newer Clang a compilation error if it's a double. It's already in the name that it's a float in case it's coming from a double (and narrowing happens).

BEGINRELEASENOTES
- Covariance matrices: Check for the size of the input for one of the constructors
  - `static_cast` the inputs to `float`

ENDRELEASENOTES


tmadlener (approved):
I will do the necessary rebasing / fixing of the README links before merging this.

PR #352, opened 2024-08-08
Remove endlines after the last include

jmcarcell:
Fix a warning that was introduced in https://github.com/AIDASoft/podio/pull/645 that appears when generating the EDM4hep files:

``` 

```

Some reestructuring (not having lines with a single ") are not needed but for consistency they are done as it is done in the rest of edm4hep.yaml.

BEGINRELEASENOTES
- Remove endlines after the last include

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -37,7 +37,6 @@ components:
         - float y
         - float z
       ExtraCode:
-        includes: "#include <cstddef>"
         declaration: "
         constexpr Vector3f() : x(0),y(0),z(0) {}\n
         constexpr Vector3f(float xx, float yy, float zz) : x(xx),y(yy),z(zz) {}\n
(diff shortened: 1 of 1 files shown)


andresailer:
Shouldn't something like this in podio
```diff
-  for include in (inc for inc in includes if inc):
+  for include in (inc for inc in includes if inc.strip()):
```

be done?

tmadlener:
I agree with Andres approach here. This should be fixed in podio. However, I can't reproduce the issue locally for some reason. Is this still happening? Resp. is it OS / python version specific?

tmadlener:
These workflows already pick up the latest version of podio as they build it on the fly. The problem in this case really seems to be that the wrong overload for `operator[](unsigned)` is chosen. Looking at #358 the writing seems to work as expected, at least the information can be read back.

tmadlener:
Obsolete with https://github.com/AIDASoft/podio/pull/670

PR #353, opened 2024-08-27
Add artifacts with EDM4hep files

jmcarcell:
BEGINRELEASENOTES
- Create a RNTuple EDM4hep file (needs podio with RNTuple) 
- Make the createED4MhepFile.py accept an argument for the output file name
- Add explanations in the README on how to obtain the files.
ENDRELEASENOTES

The artifacts have been added in this file https://github.com/key4hep/key4hep-actions/blob/main/key4hep-build/action.yml that will have to be changed once the new names here are used. With the last runs there are already some files (TTree being uploaded): https://github.com/key4hep/EDM4hep/actions/runs/10575395035


README.md
```diff
@@ -140,6 +140,14 @@ with tools that consume JSON.
 Use `edm4hep2json --help` to see the available options for selecting which parts
 of the input file should be part of the output.
 
+## Example files
+
+Example EDM4hep files can be obtained from the Continuous Integration (CI)
+workflows. From the EDM4hep github page, go to Actions -> Key4hep build, click
+one of the runs (the latest scheduled is preferred) and they will appear at the
+bottom, under Artifacts.
```
tmadlener, line 148: Are these publicly visible in the end?

m-fila: Even without github account
screenshot_2024-08-28_1

tmadlener: nice :) Thanks for checking.

m-fila:
Do I understand correctly that with `matrix` of systems and build types each job will upload an artifact and overwrite the artifact uploaded by other jobs in that batch? So the artifact that'll be available at the end would come from the job that finished last?

Also what is the retention policy for artifacts?

tmadlener:
> Do I understand correctly that with matrix of systems and build types each job will upload an artifact and overwrite the artifact uploaded by other jobs in that batch? So the artifact that'll be available at the end would come from the job that finished last?

Not entirely sure. I seem to remember that some workflows also fail if they try to overwrite artifacts from another matrix job. But that might be configurable. 

The retention policy can be configured as well. I think the default is 90 days.

README.md
peremato: Very nice. Thanks!

jmcarcell:
> Do I understand correctly that with `matrix` of systems and build types each job will upload an artifact and overwrite the artifact uploaded by other jobs in that batch? So the artifact that'll be available at the end would come from the job that finished last?

Yes, that is what would happen. This could be changed, but since we don't expect differences for different OSes or compilers if we wanted to check if they are different this should be done somewhere else.

peremato:
To be more complete. It would be nice to create files with both TTree and RNTuple.

jmcarcell:
> To be more complete. It would be nice to create files with both TTree and RNTuple.

Yes, this is what will happen once this PR is merged.

tmadlener (approved):

jmcarcell:
See https://github.com/key4hep/EDM4hep/actions/runs/10664951162 as an example, there will be a zip with a TTree file and a RNTuple file.

peremato:
Sorry @jmcarcell. Another problem. 
```
        state.d0 = next(counter)
        state.z0 = next(counter)
```
These two statements have no effect. See the TBrowser output:
Screenshot 2024-09-02 at 10 43 05

m-fila:
Looks like another capitalization typo, should be `state.D0`, `state.Z0`

jmcarcell:
Yeah, unfortunately using EDM4hep from Python has several caveats like it's possible to set any attribute even if it doesn't actually exist.

PR #354, opened 2024-08-30
Remove C++17 compatibility in version header file

m-fila:
BEGINRELEASENOTES
- Removed C++17 compatibility in version header file

ENDRELEASENOTES

Since #343 we removed compatibility with older versions of C++, this can also go as now we have the spaceship operator

### DIFF
--- a/EDM4hepVersion.h.in
+++ b/EDM4hepVersion.h.in
@@ -2,12 +2,7 @@
 #define EDM4HEP_VERSION_H
 
 #include <cstdint>
-#include <tuple>
 #include <ostream>
-#if __cplusplus >= 202002L
-#include <compare>
-#endif
-
 // Some preprocessor constants and macros for the use cases where they might be
 // necessary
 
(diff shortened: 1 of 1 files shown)


jmcarcell (commented):
While we are at this file I think https://github.com/key4hep/EDM4hep/blob/artifacts/EDM4hepVersion.h.in#L78 can be `consteval`.

tmadlener:
I am not sure if it makes sense to make the `decode_version` `consteval`. That would render it unusable in runtime contexts. But I am also not sure if we have actual use cases for that.

tmadlener (approved):
Nice. Thanks :)

jmcarcell:
> I am not sure if it makes sense to make the `decode_version` `consteval`. That would render it unusable in runtime contexts. But I am also not sure if we have actual use cases for that.

Yeah I'm not sure if there would be a case when the version would not be known at compile time, like reading it from a file...

tmadlener:
I think if they are read from file, they will be split into `major`, `minor`, `patch` instead of being encoded into the 64 bit pre processor constant. I would leave it at `constexpr` for now (as we did with this PR).

Issue #355, opened 2024-08-30
Wrong intended assignment of CovMatrix in createEDM4hepFile.py


peremato:
With the following line, the output file is written with 21 zeros instead of the intended value of `cov6f`.
https://github.com/key4hep/EDM4hep/blob/7ffc2d4ec39842e8a69cce5430f6c7489c47d517/scripts/createEDM4hepFile.py#L228

I guess it should be `state.setCovMatrix(cov6f)`

tmadlener:
I think it should also work with `state.covMatrix`

peremato:
Have a look at what is stored. If I remember correctly I was getting a covmatrix fill with zeros. See the TBrowser output:
![Uploading Screenshot 2024-09-02 at 10.43.05.png…]()

jmcarcell:
Yes, this will be fixed in https://github.com/key4hep/EDM4hep/pull/353

PR #357, opened 2024-09-04
Fix assignment to non-exisitng members in createEDM4hep.py

m-fila:
BEGINRELEASENOTES
- Fixed typos in member assignment in `createEDM4hep.py`

ENDRELEASENOTES

Fixed the typos in script creating example edm4hep file.
Due to the typos a new member was created instead of updating a value of an existing one., hence incorrect values were written to the output. Originally reported on  #353.

Since earlier there were some similar issues reported (#355) I checked if there are any more typos by setting a pythonizations that forbids dynamically creating new members:

```python
import cppyy

def freeze_class(klass, name):
    def freeze_setattr(self, attr, val):
        if not attr in type(self).__dict__:
            raise AttributeError(attr)
        old_setattr(self,attr,val)

    old_setattr = klass.__setattr__ 
    klass.__setattr__ = freeze_setattr

cppyy.py.add_pythonization(freeze_class, "edm4hep")
```

which gives errors like that:

```python
Traceback (most recent call last):
  File "/home/mafila/EDM4hep/./scripts/createEDM4hepFile.py", line 234, in <module>
    state.d0 = next(counter)
  File "/home/mafila/EDM4hep/./scripts/createEDM4hepFile.py", line 8, in freeze_setattr
    raise AttributeError(attr)
AttributeError: d0
```

The only errors were for `d0` and `z0` so hopefully that was the last bug of this type

### DIFF
--- a/scripts/createEDM4hepFile.py
+++ b/scripts/createEDM4hepFile.py
@@ -218,10 +218,10 @@ for i in range(frames):
         track.addToSubdetectorHitNumbers(next(counter))
         state = edm4hep.TrackState()
         state.location = next(counter)
-        state.d0 = next(counter)
+        state.D0 = next(counter)
         state.phi = next(counter)
         state.omega = next(counter)
-        state.z0 = next(counter)
+        state.Z0 = next(counter)
         state.tanLambda = next(counter)
         state.time = next(counter)
         state.referencePoint = edm4hep.Vector3f(
--- a/scripts/createEDM4hepFile.py
+++ b/scripts/createEDM4hepFile.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 # Description: Create a file with EDM4hep data using the EDM4hep python bindings
 # The purpose of the script is to use all the classes of the EDM4hep library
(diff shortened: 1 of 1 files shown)


m-fila:
As a side note I don't see much difference when running the script without pythonization:
```
real    0m11.938s
user    0m7.229s
sys     0m3.836s
```
and with it
```
real    0m12.041s
user    0m7.305s
sys     0m3.866s
```
(single run)

tmadlener:
Nice. The pythonization look generic enough to add it to the ones we already have available in podio, right?

jmcarcell:
Yeah, this default behaviour of being able to set anything from python is not great since mistakes like I did will always end up happening...

m-fila:
> Yeah, this default behaviour of being able to set anything from python is not great since mistakes like I did will always end up happening...

They call it a feature :wink: 

> Nice. The pythonization look generic enough to add it to the ones we already have available in podio, right?

Thank you. This version is not taking into account attributes from super class. I need to think whether it might be a problem for podio generated classes

PR #358, opened 2024-09-06
Add read tests to ensure backwards compatibility

tmadlener:
BEGINRELEASENOTES
- Download files written with EDM4hep versions for which we promise backwards compatibility and check whether we can still read them correctly.

ENDRELEASENOTES

- [x] Add infrastructure to download files on the fly
- [x] Embed datamodel version into file
  - [x] Requires https://github.com/AIDASoft/podio/pull/651 
  - [x] Requires https://github.com/AIDASoft/podio/pull/678 (for checking and future comparisons)
- [x] Needs #361 
- [x] Placement of input files after podio and EDM4hep tags have been created


tmadlener:
I will upload new files once we have tags for EDM4hep and podio, then the checks should pass. I don't foresee any more code changes now, unless there are comments from reviewing it.

tmadlener:
I have added newly produced files (from podio and EDM4hep master / main branches), to verify that things work if both are at their current head. After both have been tagged, I will upload files produced with the actually tagged version and update this accordingly.

tmadlener:
Input files from tagged podio and EDM4hep have been uploaded and are now used for the tests. The old files from intermediate testing have been removed again.

tmadlener:
Following [this comment](https://github.com/key4hep/EDM4hep/pull/361#issuecomment-2353260141), I have added `-p no:gaudi_fixtures` to the `pytest.ini`. That stops the pollution of the output with the complete python file that was used for testing. That also makes it possible to keep the normal `FAIL_REGULAR_EXPRESSION`.

Unless someone wants to still have a look and comment, I would merge this either later today or early tomorrow.

hegner:
@tmadlener - what's the issue with the failing test on the downstream-build ?

tmadlener:
That seems to be an unrelated test failure in k4SimDelphes. Looks like it has been failing there as well. I will have a look.

PR #359, opened 2024-09-08
Use the Key4hepConfig flag to set the standard, compiler flags and rpath magic.

jmcarcell:
Several settings for Key4hep packages have been being set manually in CMakeLists.txt and they differ between different packages. A new file `Key4hepConfig.cmake` with some common settings was created (https://github.com/key4hep/key4hep-dev-utils/pull/5) and pushed to several repositories and these are the cmake changes needed to use them. 

The file `Key4hepConfig.cmake` is already in the repositories, so it doesn't need to be added. It was pushed but is not being used anywhere.

BEGINRELEASENOTES
- Use the Key4hepConfig flag to set the standard, compiler flags and rpath magic.
- For EDM4hep, the standard keeps being C++20 and the rpath is set as it was before. For the compiler flags `-Wshadow` is added on top of the existing ones.

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,47 +12,14 @@ SET( ${PROJECT_NAME}_VERSION  "${${PROJECT_NAME}_VERSION_MAJOR}.${${PROJECT_NAME
 find_package(podio 1.0 REQUIRED HINTS $ENV{PODIO})
 find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics)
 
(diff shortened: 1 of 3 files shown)


tmadlener:
Can we put changes (or absence thereof) of behavior into the release notes here as well?

jmcarcell:
I added a line with the changes.

tmadlener (approved):

PR #361, opened 2024-09-10
Make sure that files with the complete datamodel can be read

tmadlener:
BEGINRELEASENOTES
- Refactor `createEDM4hepFile.py` script to facilitate testing. Apart from the obviously different values in the objects, the overall structure is almost unchanged for the output file.
  - The `MCVertexRecoParticleLinkCollection` has been renamed to the `VertexRecoParticleLinkCollection`
  - The `subdetectorHoleNumbers` are also populated for `Track`s.

ENDRELEASENOTES

I have refactored the `createEDM4hepFile.py` script to effectively reset the counter to 0 for every datatype in order to make testing a bit easier, because we won't have global changes in expected values in case some data members are added or removed. With this we can use this script to generate an EDM4hep file for every tag we have and store it somewhere (probably EOS, the same way we do it for podio) and download them on-the-fly for testing to be able to check that we don't accidentally break backwards compatibility with changes to the datamodel.

This is the first part of #358 without the backwards compatibility parts yet, to disentangle the two and to make it possible to tag this part first before adding backwards compatibility tests. In this way the latter will have a clearly reproducible path to regenerating input files if necessary.


test/backwards_compat/test_EDM4hepFile.py
```diff
@@ -0,0 +1,508 @@
+#!/usr/bin/env python3
+"""Script that can be used to check whether the file that has been created by
```
tmadlener, line 2:
```suggestion
"""Pytest module that can be used to check whether the file that has been created by
```

test/test_EDM4hepFile.py
```diff
@@ -0,0 +1,508 @@
...
+created by scripts/createEDM4hepFile.py has the expected contents
+"""
+
+import os
+import podio
+import edm4hep
+import pytest
+from itertools import count
+
+# For new simply copy these from createEDM4hepFile.py
```
jmcarcell, line 12:
```suggestion
# For now simply copy these from createEDM4hepFile.py
```

scripts/createEDM4hepFile.py
```diff
@@ -373,6 +377,124 @@
...
+        "GeneratorEventParametersCollection",
+    )
+    frame.put(create_GeneratorPdfInfoCollection(), "GeneratorPdfInfoCollection")
+
+    return frame
+
+
+if __name__ == "__main__":
+    import argparse
+    import sys
```
jmcarcell, line 475: Imports at the top, don't the python tools complain about this?

m-fila: Not configured for this repo

test/test_EDM4hepFile.py
```diff
@@ -0,0 +1,508 @@
...
+
+    check_LinkCollection(event, "CaloHitSimCaloHitLink", calo_hit, simcalo_hit)
+    check_LinkCollection(
+        event, "TrackerHitSimTrackerHitLink", tracker_hit, simtracker_hit
+    )
+    check_LinkCollection(event, "CaloHitMCParticleLink", calo_hit, particle)
+    check_LinkCollection(event, "ClusterMCParticleLink", cluster, particle)
+    check_LinkCollection(event, "TrackMCParticleLink", track, particle)
+    check_LinkCollection(event, "RecoMCParticleLink", reco_particle, particle)
+    check_LinkCollection(event, "VertexRecoParticleLink", vertex, reco_particle)
```
jmcarcell, line 508:
```suggestion
    check_LinkCollection(event, "RecoMCParticleLink", reco_particle, particle)
    check_LinkCollection(event, "CaloHitSimCaloHitLink", calo_hit, simcalo_hit)
    check_LinkCollection(
        event, "TrackerHitSimTrackerHitLink", tracker_hit, simtracker_hit
    )
    check_LinkCollection(event, "CaloHitMCParticleLink", calo_hit, particle)
    check_LinkCollection(event, "ClusterMCParticleLink", cluster, particle)
    check_LinkCollection(event, "TrackMCParticleLink", track, particle)
    check_LinkCollection(event, "VertexRecoParticleLink", vertex, reco_particle)
```
Just like the rest it's easier to check that nothing is missing if it's sorted like in `edm4hep.yaml`

scripts/createEDM4hepFile.py
```diff
@@ -6,51 +6,14 @@
...
-    cov4f = edm4hep.CovMatrix4f()
-    for i in range(10):
-        cov4f[i] = next(counter)
-    cov6f = edm4hep.CovMatrix6f()
-    for i in range(21):
-        cov6f[i] = next(counter)
 
+def create_EventHeaderCollection(vectorsize):
+    """Create an EventHeaderCollection"""
+    counter = count()
```
jmcarcell, line 16: Something that changes since now every function resets the counter is that every first field is 0 in every class which is most of the time indistinguishable from default initialization. This can be changed by changing `count()` to `count(1)` or whatever other number.

tmadlener: Good point. I have added a new constant and made all `count`s start from there.

test/test_EDM4hepFile.py
tmadlener: Done.

jmcarcell (approved):
For this PR I don't have any more comments, I'll think about the other PR related to backwards compatibility.

scripts/createEDM4hepFile.py
```diff
@@ -137,9 +128,28 @@
...
+def create_CovMatrixNf(n_dim):
+    """Create a covariance matrix of dimension n_dim"""
+    if n_dim not in (2, 3, 4, 6):
+        return ValueError(
+            f"{n_dim} is not a valid dimension for a CovMatrix in EDM4hep. Valid: (2, 3, 4, 6)"
+        )
+    counter = count(COUNT_START)
+    # With the current version of cppyy (from ROOT 6.30.06)
+    # it's not possible to initialize an std::array from a list
+    # In future versions (works with 6.32.02) it will be possible
```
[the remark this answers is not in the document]
tmadlener: Yes.

tmadlener:
It looks like `GaudiTesting` produces some `<DartMeasurement ...` XML output that contains the complete test file as a string, which in turn matches the `FAIL_REGULAR_EXPRESSION` from the test environment. I have disabled this matching for the new tests. Hopefully now everything that we expect to pass should also pass.

jmcarcell:
Hmm locally I get that is loading GaudiTesting but the tests pass. The last change here doesn't change anything for me, the tests always pass. To remove `GaudiTesting` from the plugins one can add `-p no:gaudi_fixtures -p no:collect_for_ctest -p no:ctest_measurements_reporter` to the configuration or to the command. Maybe it's worth to add it but there is some ongoing work for tests right now in Gaudi so these things may change.

tmadlener:
The tests were always passing. However, the output contained things like `getEnergyError` (from the test code) which trigger this regular expression
https://github.com/key4hep/EDM4hep/blob/6d785eb9fd91933ccf3f551477520ab2ef70c6d8/test/CMakeLists.txt#L12

PR #363, opened 2024-09-13
Add missing attributes for python package

m-fila:
BEGINRELEASENOTES
- Added missing attributes for python package

ENDRELEASENOTES

The package attributes from the import system were missing since the `edm4hep` is rebound to the `edm4hep` submodule created with ROOT/cppyy. Similarly `__version__` was set in the `__init__.py` but then lost

Having missing attributes shouldn't be a problem as we didn't experience it yet. But I think there was an intention to have `edm4hep.__version__` available

Import system [docs](https://docs.python.org/3/reference/import.html#import-related-module-attributes)

### DIFF
--- a/test/test_module.py
+++ b/test/test_module.py
@@ -1,0 +1,8 @@
+import edm4hep
+
+assert edm4hep.__version__
+assert edm4hep.__name__ == "edm4hep"
+assert edm4hep.__spec__
+assert edm4hep.__path__
+assert edm4hep.__file__
+assert "utils" in dir(edm4hep)
--- a/test/test_module.py
+++ b/test/test_module.py
@@ -1,8 +1,29 @@
-import edm4hep
+#!/usr/bin/env python3
 
-assert edm4hep.__version__
-assert edm4hep.__name__ == "edm4hep"
-assert edm4hep.__spec__
-assert edm4hep.__path__
-assert edm4hep.__file__
+"""Check the attributes of edm4hep package"""
+
+import edm4hep
+from pathlib import Path
+from ROOT import gInterpreter
+
+if gInterpreter.LoadFile("edm4hep/EDM4hepVersion.h") != 0:
+    raise ImportError("Error while loading file edm4hep/EDM4hepVersion.h")
+
+# __version__
+version = edm4hep.version.build_version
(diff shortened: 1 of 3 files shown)


test/test_module.py
```diff
@@ -0,0 +1,8 @@
+import edm4hep
+
+assert edm4hep.__version__
+assert edm4hep.__name__ == "edm4hep"
+assert edm4hep.__spec__
+assert edm4hep.__path__
+assert edm4hep.__file__
```
tmadlener, line 7: Can we easily check if all of them point to the correct / expected value?

tmadlener (commented):
Do we have to do a similar thing to podio?

m-fila:
> Do we have to do a similar thing to podio?

No, podio is fine. In podio we use in wrappers or bring elements from the cppyy created module to the `__init__.py`. 

Here we instruct the system to put under the name `edm4hep` the cppyy created module (normaly would be called `edm4hep.edm4hep`) instead of the `edm4hep` module from `edm4hep/__inity__.py`
https://github.com/key4hep/EDM4hep/blob/d7a209fd6498d983008346d30e13567564827db9/python/edm4hep/__init__.py#L31

test/test_module.py
```diff
@@ -0,0 +1,29 @@
...
+assert edm4hep.__version__ == f"{version.major}.{version.minor}.{version.patch}"
+# __name__
+name = "edm4hep"
+assert edm4hep.__name__ == name
+# __file__
+expected_origin = str(Path(__file__).parent.parent / "python" / "edm4hep" / "__init__.py")
+assert edm4hep.__file__ == expected_origin
+# __path__
+expected_path = [str(Path(expected_origin).parent)]
+assert edm4hep.__path__ == expected_path
```
tmadlener, line 23: These will obviously be different after installation, but I don't see an easy way to check them without doing an install first. Anyhow, there is no reason to expect this to not work after installation, unless I am missing something?

tmadlener (approved):

test/test_module.py
m-fila: I didn't think about it. I think it should be possible to find package without importing it and then check if the attributes' values are the same on the imported module. That should work both in build environment and after install

PR #367, opened 2024-09-16
Fix the HepPDT related cmake warning and make sure that all tests build

tmadlener:
BEGINRELEASENOTES
- Fix the cmake warning related to `HepPDT`
- Make sure that the tests that are enabled by this actually build and run with a suitable HepMC3 version

ENDRELEASENOTES


CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ include(cmake/Key4hepConfig.cmake)
 # - Use GNU-style hierarchy for installing build products
 include(GNUInstallDirs)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
tmadlener, line 21: @andresailer anything better that we can do here?

PR #370, opened 2024-09-17
add formatting and linting python in pre-commit

### LINKED ISSUE #364: Add python linters to pre-commit
Imports at the top, don't the python tools complain about this?

_Originally posted by @jmcarcell in https://github.com/key4hep/EDM4hep/pull/361#discussion_r1758530608_
            

To flag things like this automatically

m-fila:
BEGINRELEASENOTES
- Added ruff pre-commit hook for formatting and linting python files
- Reformatted and fixed warnings in python files
- Added cmake targets `ruff` and `ruff-format` to lint and format python code

ENDRELEASENOTES

Linting and formatting done with [ruff](https://docs.astral.sh/ruff/) which should be compatible with black, flake8 and pylint while having a single configuration file. ruff is already in the stack

CI added as local hook

Selected [rules](https://docs.astral.sh/ruff/rules/) `["F", "E", "W", "PLE", "PLW", "PLC"]` from flake8 (pyflakes "F", pycodestyle errors "E" and warnings "W") and pylint (errors "PLE", warnings "PLW" and conventions "PLC")
I'd also consider adding some of the `D1` checks for docstrings but there seems to be only a few docstrings now :upside_down_face: 

The codebase seems to use mixed 2-space and 4-space indent, now the default 4-space is used in formatter

Should I also add cmake targets as in #257?

Closes #364

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,13 @@ repos:
         entry: python ./scripts/updateReadmeLinks.py --check
         files: (^README.md$)
         language: system
+      - id: ruff
+        name: ruff
+        entry: ruff check --force-exclude
+        types: [python]
+        language: system
+      - id: ruff-format
+        name: ruff-format
+        entry: ruff format --force-exclude
+        types: [python]
+        language: system
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,5 +1,13 @@
 target-version = "py310"
+
 line-length = 99
+
+[format]
+# Make things format the same way as black
+quote-style = "double"
+indent-style = "space"
+skip-magic-trailing-comma = false
+line-ending = "auto"
 
 [lint]
 select = ["F", "E", "W", "PLE", "PLW", "PLC"]
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,0 +1,5 @@
+target-version = "py310"
+line-length = 99
+
+[lint]
+select = ["F", "E", "W", "PLE", "PLW", "PLC"]
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,4 +1,4 @@
-target-version = "py310"
+target-version = "py311"
 
 line-length = 99
 
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -32,7 +32,7 @@ jobs:
           cd $STARTDIR
           python3 -m venv /root/pre-commit-venv
           source /root/pre-commit-venv/bin/activate
-          pip install pre-commit
+          pip install pre-commit ruff
           cmake -DENABLE_SIO=ON \
             -DCMAKE_CXX_STANDARD=20 \
             -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror "\
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,3 +73,6 @@ add_subdirectory(test)
 
 #--- create uninstall target ---------------------------------------------------
 include(cmake/EDM4HEPUninstall.cmake)
+
+#--- code format targets -------------------------------------------------------
+include(cmake/pythonFormat.cmake)
--- a/cmake/pythonFormat.cmake
+++ b/cmake/pythonFormat.cmake
@@ -1,0 +1,25 @@
+# Additional target to run python linters and formatters on python files
+#
+# Requires ruff to be available in the environment
+
+# Get all our Python files
+file(GLOB_RECURSE ALL_PYTHON_FILES ${PROJECT_SOURCE_DIR}/python/*.py
+  ${PROJECT_SOURCE_DIR}/scripts/*.py ${PROJECT_SOURCE_DIR}/test/*.py)
+
+find_program(RUFF_EXECUTABLE ruff)
+if(RUFF_EXECUTABLE)
+    add_custom_target(
+        ruff
+        COMMAND ${RUFF_EXECUTABLE}
+        check --force-exclude ${ALL_PYTHON_FILES}
+    )
+    set_target_properties(ruff PROPERTIES EXCLUDE_FROM_ALL TRUE)
+    add_custom_target(
+        ruff-format
+        COMMAND ${RUFF_EXECUTABLE}
+        format --force-exclude ${ALL_PYTHON_FILES}
+    )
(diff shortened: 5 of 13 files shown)


m-fila:
No ruff in  the LCG releases? :eyes:

tmadlener:
> Should I also add cmake targets as in https://github.com/key4hep/EDM4hep/pull/257?

We either adapt that to run the same checks as this, or we have to include `flake8` and `pylint` in this PR as well. Alternatively, we close that and include the changes here.

.pre-commit-config.yaml
```diff
@@ -18,3 +18,13 @@ repos:
...
+      - id: ruff
+        name: ruff
+        entry: ruff check --force-exclude
+        types: [python]
+        language: system
+      - id: ruff-format
+        name: ruff-format
+        entry: ruff format --force-exclude
+        types: [python]
+        language: system
```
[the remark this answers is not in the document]
m-fila: It's supposed to be a drop-in replacement for black. They [claim](https://docs.astral.sh/ruff/formatter/#black-compatibility) to be nearly identical and running ruff on black-formatted code should give zero or minimal number of changes

I run ruff on the podio with
```
[tool.black]
line-length = 99
target-version = ["py310"]

[tool.ruff]
line-length = 99
target-version = "py310"
```

and got changes in 3 lines: removed some blank line after module docstring and these:
```diff
-        component["Component"]["ExtraCode"][
-            "const_declaration"
-        ] = "// not even valid c++ passes here"
+        component["Component"]["ExtraCode"]["const_declaration"] = (
+            "// not even valid c++ passes here"
+        )
...
-        datatype["DataType"]["ExtraCode"][
-            "invalid_extracode"
-        ] = "an invalid entry to the ExtraCode"
+        datatype["DataType"]["ExtraCode"]["invalid_extracode"] = (
+            "an invalid entry to the ExtraCode"
+        )
```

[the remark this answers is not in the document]
tmadlener: Thanks for checking. Good enough for me in EDM4hep :)

m-fila:
Added cmake targets here so #257 can be closed if this is merged

tmadlener:
Very nice. Thanks :)

tmadlener (approved):

jmcarcell:
What about including formatting as part of `Key4hepConfig.cmake`? For python ruff is very fast so it won't change much, `clang-format` for every file I don't know :thinking:

tmadlener:
I like the idea. We would probably then also need a pre-commit setup that checks in all repositories. In the end the targets are only defined but not run by default, so it would be a somewhat easy way for users to fix formatting issues that are flagged by pre-commit.

m-fila:
Hmm pre-commit config is just a single file and if we want to centrally manage it then the projects are not supposed to modify it, right?
So if we decide that we don't want for instance `clang-tidy` in pre-commit everywhere then adding `clang-tidy` just to selected repos would be a problem?
Similar problem with ruff config, ruff format and lint are in the same file so if we want to have managed ruff config with just format then it wont be possible to add lint to it. Also the config then should be probably moved from recommended `pyproject.toml` to ruff specific otherwise it won't be possible to modify `pyproject.toml`?

tmadlener:
I think some of the configuration files simply can't be managed centrally, or at least not for all packages. It's not entirely clear to me where to draw the line, but e.g. `pyproject.toml` is not something I would manage centrally, since there is a lot of project specific metadata that goes into there.

m-fila:
So should we centrally manage just the cmake target with ruff format but no linting target, pre-commit config, ruff config? Ruff can run on the defaults without any config file, the only difference between the defaults and the config here is the `target-version = "py38"` vs `target-version = "py310"`

tmadlener:
Maybe python related targets need to go into a separate `.cmake` file? I am not entirely sure if all the repositories will need it in the end. Additionally, Gaudi option files are not easily lintable, I think, since most of the things are dynamically populated. 

I would leave the pre-commit config on a per repository basis at least. (The github action can then again be distributed centrally). For the linting, I am not entirely sure, in principle we should apply the same rules for all repositories. However, that would then mean that we would need to go the way of ruff specific config files, right? Because `pyproject.toml` is definitely repository specific, IMHO.

m-fila:
I think there is no hurry for this one
I just realized I missed there is a black config in k4FWCore with different line width (99). I think I'll change it here to 99 too for some consistency

.ruff.toml
```diff
@@ -0,0 +1,5 @@
+target-version = "py310"
+line-length = 99
```
tmadlener, line 2: I was just about to say we should add this. Nice :)

tmadlener (approved):
Looks good to me, thanks. If nobody complains, I will merge this later today.

.ruff.toml
andresailer: Should this be py311? cf. https://github.com/key4hep/k4geo/pull/398/files#diff-b3236d4cb260ca6acce083493fcdc42816f732fc3d5aefb263a3c3fe9db4f81f

andresailer:
> No ruff in the LCG releases? 👀

There will in the next one: https://lcginfo.cern.ch/pkg/ruff/

(Hopefully happening Monday)

.ruff.toml
m-fila: I tried to be compatible with the black config from k4FWCore https://github.com/key4hep/k4FWCore/blob/main/pyproject.toml

andresailer: Then let's put py310 in k4geo (@tmadlener)?

m-fila: Updating to `py311` here and in k4FWCore and reformatting doesn't change any files. We have python 3.11 in the nightlies. I think we could just update and have py311 everywhere

I can also copy the extra formatting options from k4geo (it's just being explicitly setting the values same as the default right?)

tmadlener: > it's just being explicitly setting the values same as the default right?

Yes. Specifically the defaults that make ruff work / format the same as black.

tmadlener:
We could also switch over to use a key4hep stack, since that is what we are doing on other repositories, I think. Using an LCG stack was done historically, because we could more easily get an LLVM based stack (for `clang-format`), but that is now also available from a key4hep stack

.ruff.toml
tmadlener: > I think we could just update and have py311 everywhere

I would be in favor of that.

m-fila: Updated here

PR #371, opened 2024-09-17
Relax the requirement for the HepMC3 version

jmcarcell:
Introduced in https://github.com/key4hep/EDM4hep/pull/367, it shouldn't be necessary to require any version of HepMC3, see        https://github.com/key4hep/EDM4hep/commit/a98fad09172cf88e14eb7daa4ad37cca2d786819#r146821010

BEGINRELEASENOTES
- Relax the requirement for the HepMC3 version

ENDRELEASENOTES

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -104,7 +104,7 @@ IF(TARGET ROOT::ROOTDataFrame)
 endif()
 
 
-find_package(HepMC3 3.3)
+find_package(HepMC3)
 find_package(HepPDT)
 
 if(HepMC3_FOUND AND HepPDT_FOUND )


test/CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ IF(TARGET ROOT::ROOTDataFrame)
 endif()
 
 
-find_package(HepMC3 3.3)
+find_package(HepMC3)
```
tmadlener, line 107: HepMC3 implies `3.0`, I assume?

jmcarcell: Almost, there are some alpha and beta versions from 8 years ago but I don't think those are easily installable... so it should be fine
https://gitlab.cern.ch/hepmc/HepMC3/-/tags

PR #373, opened 2024-09-19
Switch to templated links and remove explicit declarations from yaml file

tmadlener:
BEGINRELEASENOTES
- Switch to templated `podio::Link`s for the `Link` datatypes
- Bump the schema version to 3
- **Remove the legacy `Association` headers that were put in place for a smooth transition to `Link`s**

ENDRELEASENOTES

This PR introduces all the existing `Link` datatypes as typedefs (keeping the same header file names for the collections as before) of `podio::LinkCollection<ToT, FromT>` and removes their declarations from the yaml file. It also includes the necessary schema evolution code for podio to be able to read back data files that have been written with previous versions of EDM4hep (or at least they will once suitable files have been uploaded for testing). 

- The typedef headers is in `include/edm4hep`
- The schema evolution code is in `edm4hep/schema_evolution`

Both of these can obviously be changed to something more appropriate if necessary.

- [x] Requires AIDASoft/podio#257
- [x] Requires AIDASoft/podio#691
- [x] Release notes with more information
- [x] Includes #358 
- [x] Updates to the README and documentation for declaring how Links are to be used (covered by podio docs)


tmadlener:
The Key4hep release workflows are failing because they don't come with a recent enough version of podio.

tmadlener:
The downstream-build workflow seems broken and it looks like it stumbles during parsing of the PR message? I will try to build this locally to check if there are any leftover *Associations* still in the stack. Otherwise I would like to get this merged rather soon.

tmadlener:
Nothing breaks for me locally, so I think this can be merged. Waiting another day to give everyone a chance to get a comment or question in.

jmcarcell:
The downstream build should be fixed, parsing was almost removed but not quite because of some `{{ }}` expansion that was messing with the script.

I was not sure about schema evolution being necessary here but 0.99 was already in July so it's all the files produced in the last 6 months with the nightlies or in the last months with the latest release :thinking:

edm4hep/CMakeLists.txt
```diff
@@ -27,6 +27,14 @@ endif()
 PODIO_ADD_ROOT_IO_DICT(edm4hepDict edm4hep "${headers}" src/selection.xml)
 add_library(EDM4HEP::edm4hepDict ALIAS edm4hepDict )
 
+add_library(edm4hep_v2 SHARED schema_evolution/src/OldLinkEvolution.cc)
+target_include_directories(edm4hep_v2 PRIVATE
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/edm4hep/schema_evolution/include>
+)
+target_link_libraries(edm4hep_v2 PUBLIC podio::podio EDM4HEP::edm4hep)
```
tmadlener, line 34: I am not entirely sure about the naming of this library here. Maybe `v2` gives the wrong impression? Ideally this would simply be merged into `edm4hep` but the problem for that is that we still need a dedicated dictionary at the moment, because the `selection.xml` for EDM4hep is generated and we cannot simply inject the few datatypes that would be necessary into that.

andresailer: Maybe `edm4hepOLE` (OldLinkEvolution)? or `edm4hepOLD` (OldLinkData)?

tmadlener: :D that would be some nice backronym in this case. Would be OK for me, but I am not sure we can find one for all future changes.

tmadlener: Renamed it to `edm4hepOldSchemas` which should hopefully be a somewhat future proof catch-all.

PR #377, opened 2024-10-09
Remove redundant inline

jmcarcell:
BEGINRELEASENOTES
- Remove redundant `inline` from templated utility functions.

ENDRELEASENOTES


tmadlener:
In this case it's the templates that (implicitly) make these `inline`, not the `constexpr`, I think. (At least that is what I got from the discussion in https://github.com/AIDASoft/podio/pull/694. 

However, it doesn't really matter, the `inline` is still redundant ;)

PR #378, opened 2024-10-23
Make sure that bit numbers are different for tertiary vertices

tmadlener:
BEGINRELEASENOTES
- Make sure that the bits used for secondary and tertiary bits are actually different

ENDRELEASENOTES

@BrieucF thanks for the hint and spotting this.


andresailer:
```
   
  -- Configuring done (2.7s)
  CMake Error at /home/runner/work/EDM4hep/EDM4hep/install/lib64/cmake/EDM4HEP/EDM4HEPTargets.cmake:61 (set_target_properties):
    The link interface of target "EDM4HEP::edm4hep" contains:
  
      nlohmann_json::nlohmann_json
```

I think we need a `find_dependency(nlohmann_json)` in our EDM4hepConfig if we build with json

(We are running into the change from here https://github.com/root-project/root/pull/16703 where ROOT used to always do that)

tmadlener:
I have opened #379 to fix that.

andresailer (approved):

PR #379, opened 2024-10-23
Find nlohmann_json downstream if EDM4hep has support

tmadlener:
BEGINRELEASENOTES
- Add an `EDM4HEP_WITH_JSON` cmake option (default `ON`) to be explicit about whether we want to build EDM4hep with JSON support or not.
  - Properly propagate this to downstream packages.
  - If set to `OFF` the `edm4hep2json` executable will not be built.

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,8 @@ find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics)
 
 option(EDM4HEP_WITH_JSON "Whether or not to build EDM4HEP with JSON support (and edm4hep2json)" ON)
 if (EDM4HEP_WITH_JSON)
- find_package(nlohmann_json 3.10.5 REQUIRED)
+ set(EDM4HEP_NLOHMANN_JSON_VERSION 3.10.5)
+ find_package(nlohmann_json ${EDM4HEP_NLOHMANN_JSON_VERSION} REQUIRED)
 endif()
 
 include(cmake/Key4hepConfig.cmake)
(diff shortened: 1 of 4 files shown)


tmadlener:
There is a potential discussion point: Should we just build EDM4hep with JSON support unconditionally? Currently we make that dependent on whether we find `nlohmann_json`, which is not very explicit, and I would actually prefer to be explicit about that e.g. via an option.

andresailer:
> There is a potential discussion point: Should we just build EDM4hep with JSON support unconditionally? Currently we make that dependent on whether we find `nlohmann_json`, which is not very explicit, and I would actually prefer to be explicit about that e.g. via an option.

:+1:

CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ SET( ${PROJECT_NAME}_VERSION  "${${PROJECT_NAME}_VERSION_MAJOR}.${${PROJECT_NAME
 find_package(podio 1.1 REQUIRED HINTS $ENV{PODIO})
 find_package(ROOT REQUIRED COMPONENTS RIO Tree Physics)
 
+option(EDM4HEP_WITH_JSON "Whether or not to build EDM4HEP with JSON support (and edm4hep2json)" ON)
+if (EDM4HEP_WITH_JSON)
+ find_package(nlohmann_json 3.10.5 REQUIRED)
```
andresailer, line 17:
```suggestion
 set(EDM4HEP_NLOHMANN_JSON_VERSION 3.10.5)
 find_package(nlohmann_json ${EDM4HEP_NLOHMANN_JSON_VERSION} REQUIRED)
```

andresailer (approved):
Thanks for putting this in so quickly!

Issue #147, opened 2022-03-15
SimCalorimeterHit and CaloHitContribution: missing fields used in EIC production


wdconinc:
Two points for discussion in calorimeter hit datatypes:
- SimCalorimeterHit does not have a time field (that info is only available in the CaloHitContributions, which are hidden behind a relation that is hard to navigate outside ROOT), however this information is needed for digitization and a fairly pertinent quantity for a SimCalorimeterHit. This adds the need for an expensive CaloHitContributions loop only to find the min(time). Other quantities technically already in CaloHitContributions (i.e. energy) are provided in SimCalorimeterHit.
- For Birks' constant saturation treatment outside of repeated geant4 running (and during digitization), CaloHitContribution would benefit from a length field to allow applying this correction. (https://en.wikipedia.org/wiki/Birks'_law). Our transition to edm4hep required us to ditch the digitization treatment of Birks' saturation.

tmadlener:
I think we are in principle fairly open to these suggestions, and more detailed discussions can probably be done in a corresponding pull request (or several to keep things separated if necessary). The main thing that currently makes us a bit reluctant in merging these already is the missing schema evolution that we are still working on in podio.

dntounis:
Hi, just to follow up on this, is there a prescription on how to assign time for SimCalorimeterHit from the time stamp of the corresponding SimCalorimeterHitContributions? I would like to study the time distribution of calorimeter hits but I'm not sure how exactly to retrieve this from the time of the HitContributions.

tmadlener:
Which time to use depends quite strongly on simulation details, resp. assumptions about how the simulated energy deposits will be readout and digitized in the detector. A common approach at the moment is to use the minimal time of all contributions as the time of the calorimeter hit they are attached to.

dntounis:
Thanks for the explanation @tmadlener !

wdconinc:
We're even going a step further (since our readout is more like a digitizer than an integrating ADC) and collecting SimCalorimeterHits with their CaloHitContributions into time-resolved SimCalorimeterPulse objects that then allow for optical signal attenuation in scintillating fiber and linear addition of multiple fibers in a bundle before they enter a SiPM. This SimCalorimeterPulse is then digitized with time-over-threshold and time-of-arrival.

PR #380, opened 2024-10-25
Added SimDRCalorimeterHit for dual-readout

wonyongc:
BEGINRELEASENOTES
- Added SimDRCalorimeterHit for optical photon dual-readout. Records number of scintillation/Cerenkov photons per hit as well as the wavelength/timing bins.

ENDRELEASENOTES


wonyongc:
Hi,
I have been developing a full simulation for a segmented crystal ECAL (CalVision/MAXICC collaborations) and am now submitting PRs to get the simulation into key4hep.
I needed to make a new datatype to store the scintllation/cerenkov hit information. So far I have simply extended the edm4hep schema with a custom 'edm4dr' schema in the simulation repo, but to merge into k4geo, it seems like it would be a better option (or the only one) to have this new schema merged into the main edm4hep.
Please let me know if I should take a different approach.
Thanks,
Wonyong

edm4hep.yaml
```diff
@@ -402,6 +402,28 @@ datatypes:
...
+    Description: "Simulated dual-readout calorimeter hit with S/C counts and wavelength/timing bins"
+    Author: "Wonyong Chung"
+    Members:
+      - uint64_t cellID                              // detector cellID
+      - float energy [GeV]                           // energy of the hit
+      - edm4hep::Vector3f position [mm]              // position of the calorimeter cell in world coords
+      - int32_t eta                                  // detector cell eta
+      - int32_t phi                                  // detector cell phi
+      - int32_t depth                                // detector cell depth
+      - int32_t system                               // detector cell system
```
andresailer, line 416: Why do you need these when there is already cellID?

edm4hep.yaml
```diff
@@ -402,6 +402,28 @@ datatypes:
...
+      - int32_t ncerenkov                            // number of cerenkov hits
+      - int32_t nscintillator                        // number of scint hits
+      - std::array<int32_t, 6000> nwavelen_cer       // number of cerenkov wavelength hits
+      - std::array<int32_t, 6000> nwavelen_scint     // number of scint wavelength hits
+      - std::array<int32_t, 6000> ntime_cer          // number of cerenkov time hits
+      - std::array<int32_t, 6000> ntime_scint        // number of scint hits
```
andresailer, line 422: What is stored in these arrays? What is the index? Why 6000?
This feels awfully inefficient.

BrieucF:
Hi Wonyong, thanks.

Do I understand correctly that, for DRCrystals, SimCalorimeter hits are not an option? Is it because, unlike for DRFibers (see https://github.com/key4hep/k4geo/pull/411), both C and S photons are produced by the same sensitive volume? 

Is the need for storing wavelength coming from the fact that you need it for wavelength dependent effects in the digitization?

SanghyunKo:
I was wondering whether we could just use two collections (one for the scintillation, one for the cherenkov channel) instead of implementing the dedicated data format (though I also opt to include the time information to the `SimCalorimeterHit` on par with the `RawCalorimeterHit`)

tmadlener:
> I was wondering whether we could just use two collections (one for the scintillation, one for the cherenkov channel)

If that would be possible then I think that would be the easiest solution. Do you already have some digitization code, where it could be exercises how workable that solution actually is there?

> (though I also opt to include the time information to the `SimCalorimeterHit`

We already discussed adding a `time` field to the `SimCalorimeterHit` in https://github.com/key4hep/EDM4hep/issues/147 without coming to a real conclusion in the end. 

The main open question at that point was, what should be stored in the `time` field for the `SimCalorimeterHit`? Since there are probably several contributions to one hit, it's not entirely obvious which one of those it should be. If there is a choice that works for the majority of cases then we could maybe add that and clearly document which time should be stored there. However, that will also mean that the sensitive action (or the EDM4hep output?) would be responsible for filling that information appropriately.

edm4hep.yaml
```diff
@@ -402,6 +402,21 @@ datatypes:
...
 
+
+  edm4hep::SimDRCalorimeterHit:
+    Description: "Simulated dual-readout calorimeter hit"
+    Author: "Wonyong Chung"
+    Members:
+      - uint64_t cellID                             // detector cellID
+      - float energy [GeV]                          // energy of the hit
+      - edm4hep::Vector3f position [mm]             // position of the calorimeter cell in world coords
+      - int32_t nCerenkovProd                       // number of cerenkov photons produced
```
andresailer, line 413: Why produced photons? Why not recorded photons? Similar to having the avg arrival time of photons recorded in the time fields?

BrieucF: This is a SIM level object, I guess we should leave the possibility to treat effects such as efficiency in a subsequent digitizer (though it is indeed sometimes interesting to leak part of the digitization in the SDAction)

andresailer: But then the tAvg should also come from the digitiser and not be calculated here already?

marco-toli: Some input from side. Our idea here is to have a first functional simulation which is a sufficiently good representation of the detector performance. The definition of the technology is still being matter of R&D and thus we are not ready yet to work on the digitization part. At this first stage, in our opinion, the most efficient  approach (also from the computational time point of view) is to factorize the simulation of energy deposition part from that of optical photon tracing, SiPM and electronics. 
From the former we need:
- number of cherenkov photons produced (we can then apply offline, at the reconstruction level a scaling on this number which takes into account the light collection efficiency, i.e. light tracing aspects)
- amount of energy deposited in each calorimetric cell

For what concerns the time stamp it is often sufficient to save the average time of all hits inside a certain calorimetric cell (possibly weighed with the energy of each hit). This is for our calorimeter a sufficiently good approximation of the time information we can extract with a realistic system (possibly after applying a resolution smearing).

It is to early to have a realistic implementation of the digitizer step in my opinion because of two reasons:
- we want to explore different configurations and for this it is more convenient to apply "digitization" effects offline
- simulation of optical photons is very time consuming so we are looking into a way to parametrize all optical aspects as much as possible (based on standalone Geant4 studies)

Concerning having separate collections for the scintillation and cherenkov photons this is also possible I believe. However, the difference with respect to the dual-readout fiber calorimeter is that in our case we have a complete duplications of hits since the cherenkov photons are always produced in the same volumes as that of other hits. Having two separate collections looked like redundant (I don't know about the impact on data size). For each hit in a certain calorimetric cell we always have both scintillation and cherenkov non-zero signals. If we use two separate collections we will have to match one-by-one all hits at the reconstruction level.

andresailer: > we want to explore different configurations and for this it is more convenient to apply "digitization" effects offline

This points towards not applying assumptions in the simulation step, but the datatype as it is now applies these assumptions for tAvg for example.

PR #381, opened 2024-11-07
Make sure to only run tests with compatible root versions

tmadlener:
BEGINRELEASENOTES
- Make sure to only run backwards compatibility tests if we have a suitable version of ROOT to still read the RC2 RNTuples stored in the files.

ENDRELEASENOTES

Necessary to make the full CI suite pass again for https://github.com/AIDASoft/podio/pull/702


tmadlener:
@andresailer @jmcarcell Do you consider this an acceptable solution to unblock the podio CI for now?

Issue #382, opened 2024-11-21
Rawhits removed from TrackerHits


andresailer:
> * Remove the edm4hep::ObjectID from the components #260

This was not quite the descriptive PR title that I thought it was, and it is turning out to be quite annoying to deal with for us due to the removal of the vector member of raw hits in TrackerHit. Same for ACTS, which is [relying](https://github.com/acts-project/acts/blob/v37.4.0/Examples/Io/EDM4hep/src/EDM4hepUtil.cpp#L176) on the raw hits. *Edit:* ACTS abuses this somewhat, but we are using this in the intended way: to refer to raw hits.

_Originally posted by @wdconinc in https://github.com/key4hep/EDM4hep/issues/261#issuecomment-2489826347_

tmadlener:
Hi @wdconinc, thanks for the report. That was definitely an unintended consequence of that change. The main goal was to get rid of the duplication of `podio::ObjectID` and `edm4hep::ObjectID`.

A few questions to see how we can salvage this best:
- What are the raw hits in your case? And how many different types are involved?
- How does your workflow look at the moment? I.e. how do you get from the ObjectID to the actual thing in the end? Did you write some utility for that?
- Since the `edm4hep::Track` now uses the `edm4hep::TrackerHit` interface for its `hits` relation, would you need access to the raw hits from the interface, or are you OK with down-casting to the actual type and getting access to the raw hits from there?
- (Could the raw hits be combined into an interface themselves? Or do you think that would be more cumbersome than the current workflow?)

wdconinc:
> Hi @wdconinc, thanks for the report. That was definitely an unintended consequence of that change. The main goal was to get rid of the duplication of `podio::ObjectID` and `edm4hep::ObjectID`.

Yeah, and that main goal is what I got from the PR title, and I agree its duplication that should be removed. I was surprised when it turned out to include other changes that I didn't expect.

> 
> A few questions to see how we can salvage this best:
> - What are the raw hits in your case? And how many different types are involved?

We are using an EDM4eic RawTrackerHit since there is no similar type available in EDM4hep. https://github.com/eic/EDM4eic/blob/09d18ef63f8b18ecc94ce995d812ab90d597acc0/edm4eic.yaml#L351. We have an EDM4eic TrackerHit too which we use in some subsystes but not all, and we would prefer to move all to EDM4hep instead of the other way round.

[...]

PR #385, opened 2024-12-02
Drift chamber digi

BrieucF:
BEGINRELEASENOTES
- Add a data type for Drift Chamber digitized hits (before the left right ambiguity is resolved)

ENDRELEASENOTES

Tagging @atolosadelgado

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -503,6 +503,22 @@ datatypes:
       void setCovMatrix(float value, edm4hep::Cartesian dimI, edm4hep::Cartesian dimJ) { getCovMatrix().setValue(value, dimI, dimJ); }\n
       "
 
+  edm4hep::DriftChamberDigi:
+    Description: "Digitized drift chamber hit (before tracking)"
+    Author: "EDM4hep authors"
+    Members:
+      - uint64_t cellID // ID of the sensor that created this hit
+      - int32_t type // type of the raw data hit
+      - int32_t quality // quality bit flag of the hit
+      - float time [ns] // time of the hit
+      - float eDep [GeV] // energy deposited on the hit
+      - float eDepError [GeV] // error measured on eDep
+      - edm4hep::Vector3d position [mm] // point on the sensitive wire (SW) which is closest to the simhit
+      - edm4hep::Vector3d directionSW // direction of SW
+      - float distanceToWire [mm] // distance hit-wire
+      - uint32_t nCluster // number of clusters associated to this hit
+    VectorMembers:
+      - uint16_t nElectrons // number of electrons for each cluster (vector size = nCluster)
 
   edm4hep::RawTimeSeries:
     Description: "Raw data of a detector readout"
@@ -809,6 +825,33 @@ datatypes:
       void {name}::setRec(const edm4hep::ReconstructedParticle& rec) { setTo(rec); }\n
       "
 
+  edm4hep::DriftChamberDigiSimTrackerHitLink:
+    Description: "Link between a DriftChamberDigi and a SimTrackerHit"
+    Author: "EDM4hep authors"
+    Members:
+      - float weight                      // weight of this link
+    OneToOneRelations:
+     - edm4hep::DriftChamberDigi from     // reference to the DriftChamberDigi
+     - edm4hep::SimTrackerHit to          // reference to the SimTrackerHit
+    ExtraCode:
+      declaration: "
+      [[deprecated(\"use getFrom instead\")]] edm4hep::DriftChamberDigi getRec() const;\n
+      [[deprecated(\"use getTo instead\")]] edm4hep::SimTrackerHit getSim() const;\n"
+      implementation: "
+      edm4hep::DriftChamberDigi {name}::getRec() const { return getFrom(); }\n
+      edm4hep::SimTrackerHit {name}::getSim() const { return getTo(); }\n"
+    MutableExtraCode:
+      declaration: "
+      [[deprecated(\"use setFrom instead\")]]\n
+      void setRec(const edm4hep::DriftChamberDigi& rec);\n
+      [[deprecated(\"use setTo instead\")]]\n
+      void setSim(const edm4hep::SimTrackerHit& sim);\n
+      "
+      implementation: "
+      void {name}::setRec(const edm4hep::DriftChamberDigi& digi) { setFrom(digi); }\n
+      void {name}::setSim(const edm4hep::SimTrackerHit& simHit) { setTo(simHit); }\n
+      "
+
 
   edm4hep::TimeSeries:
     Description: "Calibrated Detector Data"
@@ -875,3 +918,4 @@ interfaces:
     Types:
       - edm4hep::TrackerHit3D
       - edm4hep::TrackerHitPlane
+      - edm4hep::DriftChamberDigi
--- a/README.md
+++ b/README.md
@@ -25,30 +25,30 @@ A generic event data model for future HEP collider experiments.
 | [EventHeader](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L242)         | [MCParticle](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L254)        | [SimTrackerHit](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L321)         |
 | [CaloHitContribution](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L362) | [SimCalorimeterHit](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L374) | [RawCalorimeterHit](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L385)     |
 | [CalorimeterHit](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L394)      | [ParticleID](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L405)        | [Cluster](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L419)               |
-| [TrackerHit3D](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L451)          | [TrackerHitPlane](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L477)   | [RawTimeSeries](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L507)                |
-| [Track](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L521)               | [Vertex](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L538)            | [ReconstructedParticle](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L584) |
-| [TimeSeries](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L813) | [RecDqdx](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L825) | [DriftChamberDigi](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L506) |
+| [TrackerHit3D](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L451)          | [TrackerHitPlane](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L477)   | [RawTimeSeries](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L523)                |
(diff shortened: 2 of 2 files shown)


edm4hep.yaml
```diff
@@ -503,6 +503,22 @@ datatypes:
...
+    Description: "Digitized drift chamber hit (before tracking)"
+    Author: "EDM4hep authors"
+    Members:
+      - uint64_t cellID // ID of the sensor that created this hit
+      - int32_t type // type of the raw data hit
+      - int32_t quality // quality bit flag of the hit
+      - float time [ns] // time of the hit
+      - float eDep [GeV] // energy deposited on the hit
+      - float eDepError [GeV] // error measured on eDep
+      - edm4hep::Vector3d position [mm] // point on the sensitive wire (SW) which is closest to the simhit
```
andresailer, line 516: This reads like it is using Monte Carlo information?

edm4hep.yaml
```diff
@@ -503,6 +503,22 @@ datatypes:
...
+      - edm4hep::Vector3d directionSW // direction of SW
```
andresailer, line 517:
```suggestion
      - edm4hep::Vector3d directionSW // direction of the sensitive wire (SW)
```
? Along the sensitive wire? Whose direction?

SW is like always software, so using SW should be explicit everywhere and never assume people will know this is sensitive wire.

edm4hep.yaml
```diff
@@ -503,6 +503,22 @@ datatypes:
...
+      - float distanceToWire [mm] // distance hit-wire
+      - uint32_t nCluster // number of clusters associated to this hit
```
andresailer, line 519: Is this really needed if the vector member exists anyway?
Add a helper function to return nCluster from `nElectrons.size()`?

edm4hep.yaml
```diff
@@ -809,6 +825,33 @@ datatypes:
...
+    Description: "Link between a DriftChamberDigi and a SimTrackerHit"
+    Author: "EDM4hep authors"
+    Members:
+      - float weight                      // weight of this link
+    OneToOneRelations:
+     - edm4hep::DriftChamberDigi from     // reference to the DriftChamberDigi
+     - edm4hep::SimTrackerHit to          // reference to the SimTrackerHit
+    ExtraCode:
+      declaration: "
+      [[deprecated(\"use getFrom instead\")]] edm4hep::DriftChamberDigi getRec() const;\n
```
andresailer, line 838: Why are there already deprecated functions added to a new datatype?

edm4hep.yaml
```diff
@@ -503,6 +503,22 @@ datatypes:
       void setCovMatrix(float value, edm4hep::Cartesian dimI, edm4hep::Cartesian dimJ) { getCovMatrix().setValue(value, dimI, dimJ); }\n
       "
 
+  edm4hep::DriftChamberDigi:
```
andresailer, line 506: SensitiveWireHit or DriftChamberHit? 

"DriftChamberDigi" sounds like the algorithm that does the digitisation.

edm4hep.yaml
BrieucF: SW spelled in the comment (left `directionSW` as is) and tried to clarify the meaning of this member

edm4hep.yaml
BrieucF: Done

edm4hep.yaml
BrieucF: They were there in case people wanted to use links in the "Association" way. But maybe we should not let that happen since there is no existing code using this data type. Removed these helpers.

edm4hep.yaml
BrieucF: Went for SensitiveWireHit

BrieucF:
Thanks Andre

edm4hep.yaml
```diff
@@ -503,6 +503,27 @@ datatypes:
...
+      - int32_t type // type of the raw data hit
+      - int32_t quality // quality bit flag of the hit
+      - float time [ns] // time of the hit
+      - float eDep [GeV] // energy deposited on the hit
+      - float eDepError [GeV] // error measured on eDep
+      - edm4hep::Vector3d position [mm] // point on the sensitive wire which is closest to the hit (center of the circle)
+      - edm4hep::Vector3d directionSW // direction of the sensitive wire (normal vector to the circle)
+      - float distanceToWire [mm] // distance hit-wire (radius of the circle)
+    VectorMembers:
+      - uint16_t nElectrons // number of electrons for each cluster (number of clusters = vector size)
```
tmadlener, line 520: What is the typical number of electrons per cluster? I suppose 65535 is enough to work in all cases?

edm4hep.yaml
```diff
@@ -875,3 +905,4 @@ interfaces:
     Types:
       - edm4hep::TrackerHit3D
       - edm4hep::TrackerHitPlane
+      - edm4hep::SensitiveWireHit
```
tmadlener, line 908: How confusing will the slightly different meanings of `position` be in this case? Resp. do we need to worry about this?

edm4hep.yaml
```diff
@@ -503,6 +503,27 @@ datatypes:
...
+      - float eDepError [GeV] // error measured on eDep
+      - edm4hep::Vector3d position [mm] // point on the sensitive wire which is closest to the hit (center of the circle)
+      - edm4hep::Vector3d directionSW // direction of the sensitive wire (normal vector to the circle)
+      - float distanceToWire [mm] // distance hit-wire (radius of the circle)
+    VectorMembers:
+      - uint16_t nElectrons // number of electrons for each cluster (number of clusters = vector size)
+    ExtraCode:
+      declaration: "
+        /// Return the number of clusters associated to the hit\n
+        uint32_t nClusters() const { return getNElectrons().size(); }\n
```
tmadlener, line 524: Is there a specific reason to make this a `uint32_t` instead of just going with the `size_t` (or even just `auto`) that `vector::size` returns?

tmadlener, line 524:
```suggestion
        uint32_t getNClusters() const { return getNElectrons().size(); }\n
```
To fit with the overall naming schema of getters in EDM4hep.

edm4hep.yaml
BrieucF: Yes, `uint16` will do the job, this value is typically below 10.

edm4hep.yaml
BrieucF: Yes, if used blindly together with the position of the other data types in this interface, things will go wrong (a 3D hit position is ill-defined for this new data type). We went for the less intrusive option w.r.t. existing edm4hep but we can remove the position from the interface and name it differently for the `SensitiveWireHit`, if there is a consensus that this is the way to go. Probably a good thing to discuss tomorrow?

edm4hep.yaml
BrieucF: Done

BrieucF: Went for auto

edm4hep.yaml
atolosadelgado: > What is the typical number of electrons per cluster? I suppose 65535 is enough to work in all cases?

According to the paper they published [1], Fig 15, right plot:
image
There is a tail that goes up to 450, but the probability is extremely low, most of the clusters have less than 10 electrons as Brieuc said.

[1] https://arxiv.org/pdf/2105.07064

tmadlener: 450 still fits into 9 bits, so there is a 7 bit safety margin. I think this is OK. Thanks for the reference.

edm4hep.yaml
```diff
@@ -511,11 +511,13 @@ datatypes:
...
       - float time [ns] // time of the hit
-      - float eDep [GeV] // energy deposited on the hit
-      - float eDepError [GeV] // error measured on eDep
-      - edm4hep::Vector3d position [mm] // point on the sensitive wire which is closest to the hit (center of the circle)
+      - float eDep [GeV] // energy deposited by the hit
+      - float eDepError [GeV] // error on eDep
       - edm4hep::Vector3d directionSW // direction of the sensitive wire (normal vector to the circle)
-      - float distanceToWire [mm] // distance hit-wire (radius of the circle)
+      - edm4hep::Vector3d position [mm] // point on the sensitive wire which is closest to the hit (center of the circle)
+      - double positionAlongWireError [mm] // error on the hit position along the wire direction
```
tmadlener, line 518: How would the interpretation of the `positionAlongWireError` work in this case? IIUC I would need the `directionSW` to get the direction and then the `position` to get a point on the wire to get a reference point(?).

BrieucF: Correct, the reference point on the wire is `position`, then to know "in which direction the error is defined" without accessing the geometry, you need `sensitiveWireDirection` (which I just renamed for more clarity)

BrieucF:
I have added a proposal for the errors. I also applied @gaede suggestion and made the direction vector 2D (which I also renamed to be more clear). Comments welcome

BrieucF:
Actually, this data type would be more robust with explicit fields for the wire stereo and azimuthal angle (current proposal updated to this). With the vector2f, users might have different conventions.

gaede:
As discussed this morning, you could also make this class look very similar to the `TrackerHitPlane`, e.g. by replacing 
```yaml
- edm4hep::Vector2d sensitiveWireDirection // direction of the sensitive wire (normal vector to the circle)
```
with
```yaml
  - edm4hep::Vector2f u            // measurement direction along the wire - (theta, phi) in global coordinates 
  - edm4hep::Vector2f v            // measurement direction orthogonal to the wire -(theta, phi) in global coordinates 
  - float du                       // measurement error along the wire 
  - float dv                       // measurement error of distance to the wire 
```
this way you'd have the uncertainties in the class, and we would express similar concepts with similar/same names.
Once tracking is done, you could updated `v` such that `m = pos + v` is the actual position of the hit base on the track fit...

tmadlener:
Are the `u` and `v` also labelled as *global coordinates* in the `TrackerHitPlane`? I though they are to be interpreted as local coordinates?

gaede:
The directions `u` and `v` are meant to be in global coordinates - see: [https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHitPlane.html](https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1TrackerHitPlane.html).  There seems sth. has been lost in translation when copying this over from LCIO. Should update our yaml file...

jmcarcell:
For the inclusion, if this makes it easier for people to use and test it we could add it to EDM4hep and then we can add warnings to make it clear that it is an experimental type (breaking changes will happen), and also not include it in our "official" CI files. Then, no one should be upset if the files that have collections with these are not readable after it changes in EDM4hep. We should foresee this happening after EDM4hep 1.0 and always denying adding a new type until it is completed is quite restrictive. In addition, if most (all?) of the users of the new type are here and are aware it may change then it should be easier. But I can also see a type being added as an experiment forever...

BrieucF:
> > Once tracking is done, you could updated `v` such that `m = pos + v` is the actual position of the hit base on the track fit...
> 
> No such thing as _updating after ..._ in EDM4hep. That would mean cloning + setting a new value in that clone.
> 
> I think if things are conceptually different enough then it makes sense to also keep their abstractions different. Choosing the wrong abstraction might do more damage than not abstracting enough. Since this type should represent the digitized measurement that comes out of a drift chamber, maybe we have to make sure it does that first and only as a (somewhat) secondary concern think about how it relates to tracking, resp. how it would be used in the final fitted track as a tracker hit.
> 
> With that in mind and also regarding some parts of the discussion in #382, I think we will probably have to think a bit about what a _TrackerHit_ actually is, and also what a "raw tracker hit" could be and how we can connect the two things in the end.
> 
> > not include it in our "official" CI files. Then, no one should be upset if the files that have collections with these are not readable after it changes in EDM4hep.
> 
> I think we should still write them, but potentially not check them on the reading side. The reason I say we should still write them is that we don't want to break the rest of the files simply because they contain collections of datatypes that are not yet stable. Currently podio always reads full frames with no way of just skipping certain collections (until we merge [AIDASoft/podio#504](https://github.com/AIDASoft/podio/pull/504)). In principle the buffer creation mechanism takes into account the schema version so that at least connecting the buffers to the branches should work correctly even if things _would_ break during actual schema evolution later (in case of unsupported evolutions).
> 
> Here we have to sit down and define some sort of policy on how we want to deal with this.

This is interesting and we should indeed have this discussion but it goes somewhat beyond the scope of this pull request. I am afraid we won't converge in the timescale within which this PR is needed. Shall I add 'EXPERIMENTAL' to the `Description` field for now?

edm4hep.yaml
tmadlener: I would have a preference for keeping it at the moment. However, I think we should discuss this again to see how to best make this work or if this is truly a concern that actually matters in reality.

edm4hep.yaml
```diff
@@ -486,6 +486,30 @@ datatypes:
...
+      - float distanceToWire [mm] // distance between the hit and the wire (radius of the circle)
+      - float distanceToWireError [mm] // error on distanceToWire
+    VectorMembers:
+      - uint16_t nElectrons // number of electrons for each cluster (number of clusters = vector size)
+    ExtraCode:
+      declaration: "
+        /// Return the number of clusters associated to the hit\n
+        auto getNClusters() const { return getNElectrons().size(); }\n
+        "
```
jmcarcell:
```suggestion
      declaration: |
        /// Return the number of clusters associated to the hit
        auto getNClusters() const { return getNElectrons().size(); }
```

PR #390, opened 2024-12-18
Make the PIDHandler class non-copyable

jmcarcell:
BEGINRELEASENOTES
- Make the PIDHandler class non-copyable

ENDRELEASENOTES

Sounds like a bad idea to copy all these maps, vector of strings, strings...


jmcarcell:
> Probably a good idea to prevent this. Can you add a comment that we disable this because of cost not for another technical reason?

Done

PR #397, opened 2024-12-20
add python bindings for PIDHandler

### LINKED ISSUE #396: Make `PIDHandler` available and work from python
It's currently not possible to use (or even import) the `PIDHandler` from python.

m-fila:
BEGINRELEASENOTES
- Added Python bindings for PIDHandler

ENDRELEASENOTES

Python test is selected tests from C++ version
Closes #396

### DIFF
--- a/test/utils/test_PIDHandler.py
+++ b/test/utils/test_PIDHandler.py
@@ -1,0 +1,84 @@
+import unittest
+from edm4hep import ReconstructedParticleCollection, ParticleIDCollection, ReconstructedParticle
+import edm4hep
+import cppyy
+
+
+def create_recos():
+    coll = ReconstructedParticleCollection()
+    for i in range(1, 10):
+        reco = coll.create()
+        reco.setPDG(i)
+    return coll
+
+
+def create_particle_ids(recos, likelihood):
+    coll = ParticleIDCollection()
+    for reco in recos:
+        pid = coll.create()
+        pid.setLikelihood(likelihood)
+        pid.setParticle(reco)
+        pid.addToParameters(1.23)
+        pid.addToParameters(3.14)
+    return coll
+
+
+class TestPIDHandler(unittest.TestCase):
+    def test_pid_handler_basics(self):
+        reco_coll = create_recos()
+        pid_coll1 = create_particle_ids(reco_coll, 1.0)
+        pid_coll2 = create_particle_ids(reco_coll, 2.0)
+
+        handler = edm4hep.utils.PIDHandler()
+        handler.addColl(pid_coll1)
+        handler.addColl(pid_coll2)
+
+        reco = reco_coll[0]
+        pids = handler.getPIDs(reco)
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
Ah nice. I had mainly opened that issue as a reminder for myself not as a call to action ;)

What would be nice to check as well would be the constructors from frames, or from multiple collections via the static `from` methods. However, they clash with the `from` keyword from python. That was the main reason for me to not directly go to solve this ;) Would probably need some `from_event` and `from_collections` overloads on the c++ side to make things available on the python side in a working manner.

test/utils/test_PIDHandler.py
```diff
@@ -0,0 +1,84 @@
...
+
+        handler = edm4hep.utils.PIDHandler()
+        handler.addColl(pid_coll1)
+        handler.addColl(pid_coll2)
+
+        reco = reco_coll[0]
+        pids = handler.getPIDs(reco)
+        self.assertEqual(len(pids), 2)
+        self.assertAlmostEqual(pids[0].getParticle(), reco, places=5)
+        self.assertAlmostEqual(pids[1].getParticle(), reco, places=5)
```
tmadlener, line 40:
```suggestion
        self.assertEqual(pids[0].getParticle(), reco,)
        self.assertEqual(pids[1].getParticle(), reco)
```

These should always be true (and they are not even floating point numbers but actual objects).

m-fila:
There seems to be some problems with a lifetime when using `std::optional` and obtaining values from it :neutral_face:

m-fila:
I think a possible workaround for the problems with pythonized `std::optional` would be to add throwing interface and use it python bindings (and disable optional interface there)

tmadlener:
If we are adding a dedicated python wrapper / interface here, that could also accommodate the necessary handling of the `Frame` wrapper.

m-fila:
> If we are adding a dedicated python wrapper / interface here, that could also accommodate the necessary handling of the `Frame` wrapper.

I'm not sure if I understand. The functions here are expecting to get a python wrapper for Frame and then get cppyy Frame object from it and use it with the functions expecting C++ Frame

Do you mean you'd like to have reimplementation of C++ functions in python using Frame wrapper methods without reaching to C++ Frame directly?

tmadlener:
No, I was thinking that if we want the special handling of `std::optional` for the python PIDHandler, we need to write a python wrapper in any case. But now that I look at the diff again, I see that is already there(?) with the pythonizations doing all the necessary work of reaching through to the `cppyy` `podio::Frame`.

So now the actual question is: Can we make the `std::optional` work "as expected for python" with these pythonizations as well?

m-fila:
I see. I tried hiding handling of  `std::optional` inside the pythonizations hoping that if all the checking/unwrapping is done as soon as possible the lifetime issues might not kick-in. But I didn't manage to get it working reliably

tmadlener:
Closing this for now. There seems to be no real use case for this at the moment.

PR #398, opened 2025-01-07
Make the event number 64 bits and unsigned

tmadlener:
BEGINRELEASENOTES
- Make the `EventHeader::eventNumber` a 64 bit unsigned integer to avoid issues with the range of 32 bit integers.

ENDRELEASENOTES

This is another small fix that came out of the discussion with HepMC folks (and their experience from LHC). We are likely to run into issues with the range of 32 bit integers and the exepected event numbers. LHC is going beyond 32 bit range already in long runs.


tmadlener:
This breaks backwards compatibility for RNTuple as that does not have the necessary schema evolution support to do this on the fly (at least for RC2 of RNTuple).

jmcarcell:
But I guess it's quite unlikely someone has RNTuples files and a file with event headers at the same time. So recreate the RNTuple file for version 0.99 with a 64 bit event number? It will be unreadable by EDM4hep 0.99, but is that even going to happen?

tmadlener:
> So recreate the RNTuple file for version 0.99 with a 64 bit event number?

Yes that is also what I would do. We should also go to the release RNTuple format there, i.e. ROOT >= 6.34, I think.

edm4hep.yaml
```diff
@@ -222,7 +222,7 @@ datatypes:
     Description: "Event Header. Additional parameters are assumed to go into the metadata tree."
     Author: "EDM4hep authors"
     Members:
-      - int32_t eventNumber    // event number
+      - uint64_t eventNumber    // event number
       - int32_t runNumber      // run number
```
jmcarcell, line 226: Maybe the run number should be `uint32_t`?

tmadlener: done.

jmcarcell (approved):
I think there is nothing left to do in this one. Related to this PR, maybe we could review some of the `int32_t` that are in other types in EDM4hep.

PR #401, opened 2025-01-30
Use YAML multiline strings for extra code

paulgessinger:
I propose here to use YAML's native support for multiline strings to make the extra code and declarations etc more maintainable. It allows implicitly taking newlines from the string, which allows skipping putting them in explicitly.

BEGINRELEASENOTES
- Refactored YAML to use multiline strings

ENDRELEASENOTES

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -15,20 +15,20 @@ components:
         - float t
       ExtraCode:
         includes: "#include <cstddef>"
-        declaration: "
-        constexpr Vector4f() : x(0),y(0),z(0),t(0) {}\n
-        constexpr Vector4f(float xx, float yy, float zz, float tt) : x(xx),y(yy),z(zz),t(tt) {}\n
-        constexpr Vector4f(const float* v) : x(v[0]),y(v[1]),z(v[2]),t(v[3]) {}\n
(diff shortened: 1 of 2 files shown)


tmadlener:
Nice, thanks.

To fix the links, you need to run `./scripts/updateReadmeLinks.py README.md` and it looks like there is some trailing whitespaces about which pre-commit also complains.

edm4hep.yaml
```diff
@@ -470,17 +465,50 @@ datatypes:
...
+    VectorMembers:
+      - float measurement // meas
+      - float covariance // cov
+
+    # ExtraCode:
+    #   includes: |
+    #     #include <edm4hep/Vector3d.h>
+    #   declaration: |
+    #     edm4hep::Vector3d getPosition() const { return {}; }
+    #     edm4hep::Vector3d& getPosition() const { return {}; }
```
tmadlener, line 511:
```suggestion
```
Probably not part of this PR(?).

paulgessinger: Sorry.

paulgessinger: Should be gone.

m-fila (commented):
Could we use this occasion to also get rid of remaining "\n" in `ExtraCode.includes` directive? :slightly_smiling_face:

tmadlener (approved):
AFAICT this is good to go, right?

PR #403, opened 2025-02-04
show standalone documentation pages in doxygen docs

m-fila:
BEGINRELEASENOTES
- Added a tab with standalone documentation pages in doxygen docs

ENDRELEASENOTES

Adding tabs:
-`Release notes`
-`Selected topics` - I'm not entirely convinced by this name.

I didn't figure out if there is a way to use more readable references to the files

image


tmadlener:
> I didn't figure out if there is a way to use more readable references to the files

With this you mean the `url="@ref md_doc_2_p_i_d_handler"` part? I think I had a look at some point, but also couldn't figure out a better way than this one.

I am also not sure about *Selected Topics* but haven't found a better name yet either. Let's see if we one comes up in the next few days otherwise we go with this and change it later.

tmadlener:
Merging this as is, since we couldn't come up with anything better than *Selected Topics*.

PR #404, opened 2025-02-04
[WIP] Replace the spin by a helicity property for the MCParticle

tmadlener:
BEGINRELEASENOTES
- Replace the `spin` by `helicity` in `MCParticle` as the former is not usable without the full spin-density matrix whereas the latter is always well defined (or unset).
  - Defaults to `9` which is used to signify an unset value. Use `hasHelicity` to check whether the `helicity` is set

ENDRELEASENOTES

- [ ] Schema evolution to make this a backwards compatible change, such that `spin.z` will end up in `helicity` as that seems to be the main way that `spin` has been populated in the past in LCIO.

### DIFF
--- a/edm4hep/schema_evolution/include/edm4hep/schema_evolution/OldMCParticleData.h
+++ b/edm4hep/schema_evolution/include/edm4hep/schema_evolution/OldMCParticleData.h
@@ -1,0 +1,38 @@
+#ifndef EDM4HEP_SCHEMA_EVOLUTION_OLDMCPARTICLEDATA_H
+#define EDM4HEP_SCHEMA_EVOLUTION_OLDMCPARTICLEDATA_H
+
+#include <edm4hep/Vector3d.h>
+#include <edm4hep/Vector3f.h>
+
+#include <cstdint>
+
+namespace edm4hep::v2 {
+
+struct Vector2i {
+  int a{};
+  int b{};
+};
+
+struct MCParticleData {
+  std::int32_t PDG{};             ///< PDG code of the particle
+  std::int32_t generatorStatus{}; ///< status of the particle as defined by the generator
+  std::int32_t simulatorStatus{}; ///< status of the particle from the simulation program - use BIT constants below
+  float charge{};                 ///< particle charge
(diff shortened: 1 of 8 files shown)


edm4hep/schema_evolution/include/edm4hep/schema_evolution/OldMCParticleData.h
```diff
@@ -0,0 +1,38 @@
...
+  float charge{};                 ///< particle charge
+  float time{};  ///< creation time of the particle in wrt. the event, e.g. for preassigned decays or decays in flight
+                 ///< from the simulator [ns]
+  double mass{}; ///< mass of the particle [GeV]
+  ::edm4hep::Vector3d vertex{};             ///< production vertex of the particle [mm]
+  ::edm4hep::Vector3d endpoint{};           ///< endpoint of the particle [mm]
+  ::edm4hep::Vector3d momentum{};           ///< particle 3-momentum at the production vertex [GeV]
+  ::edm4hep::Vector3d momentumAtEndpoint{}; ///< particle 3-momentum at the endpoint [GeV]
+  ::edm4hep::Vector3f spin{};               ///< particle spin
+  Vector2i colorFlow{};
```
m-fila, line 29: Should this still include the `colorFlow` and `Vector2i`that were removed earlier?

tmadlener: Yes, I think so, because the decisive factor is the schema version and for schema version 2, these were still there. But I am still debugging why this isn't work as expected yet. I have the feeling we are not tricking ROOT properly yet...

wdconinc:
Question: how are transversely polarized particles intended to be handled after this is merged? In the EIC those are a key part of the program, including things like tensor polarization.

tmadlener:
As discussed during the meeting today we will merge this without backwards compatibility (to previous `v00-99-xx` tags) yet to see if it's necessary or whether `spin` has largely been unused so far. Depending on the timeline of various related developments in podio and ROOT we should be able to restore backwards compatibility on rather short notice if necessary.

PR #414, opened 2025-03-10
use link collection `typeName`

m-fila:
BEGINRELEASENOTES
- Replace `podio::detail::linkCollTypeName` with  link collection `typeName` introduced in AIDASoft/podio#748

ENDRELEASENOTES

Replace `podio::detail::linkCollTypeName` with  link collection `typeName` introduced in AIDASoft/podio#748


tmadlener:
We will have to bump the minimum required version of podio to 1.3 after this (once that tag is available). However, that will be necessary because of other things as well.

tmadlener:
Key4hep nightlies based workflows will work again tomorrow (but we need to merge this before).

Issue #415, opened 2025-03-10
Bump minimum podio version to 1.3


tmadlener:
Some PRs will lead to a new minimal dependency on podio. Once podio 1.3 has been tagged we should bump the minimum required version here.

PRs:
- #414 

Changes that can be addressed
- #413 the test that is disabled there can be enabled again with a new input file

PR #422, opened 2025-03-25
add missing units in datamodel yaml file

m-fila:
BEGINRELEASENOTES
- Add missing units for covariances, particle charge and other members

ENDRELEASENOTES

There are some cases where a unit was missing, mostly for covariances. I also suggest to add that particle charge is measured in the units the elementary charge

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -510,7 +510,7 @@ datatypes:
     Description: "Vertex"
     Author: "EDM4hep authors"
     Members:
-      - uint32_t            type          // Flagword that defines the type of the vertex, see reserved bits for more information
+      - uint32_t            type          // flagword that defines the type of the vertex, see reserved bits for more information
       - float               chi2          // chi-squared of the vertex fit
       - int32_t             ndf           // number of degrees of freedom of the vertex fit
       - edm4hep::Vector3f   position [mm]     //  position of the vertex
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -494,7 +494,7 @@ datatypes:
     Author: "EDM4hep authors"
     Members:
       - int32_t type                         // flagword that defines the type of track
-      - float chi2                       // Chi^2 of the track fit
+      - float chi2                       // chi-squared of the track fit
       - int32_t ndf                          // number of degrees of freedom of the track fit
       - int32_t Nholes                   // number of holes on track
     VectorMembers:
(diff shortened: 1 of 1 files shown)


edm4hep.yaml
```diff
@@ -398,7 +398,7 @@ datatypes:
       - float                energy [GeV]        // energy of the cluster
       - float                energyError [GeV]   // error on the energy
       - edm4hep::Vector3f    position [mm]       // position of the cluster
-      - edm4hep::CovMatrix3f positionError   // covariance matrix of the position
+      - edm4hep::CovMatrix3f positionError [mm**2]  // covariance matrix of the position
```
andresailer, line 401: There is one prior instance of the use of ^ to indicate an exponent in the yaml file.
```Chi^2```

```suggestion
      - edm4hep::CovMatrix3f positionError [mm^2]  // covariance matrix of the position
```

m-fila: There is also one prior instance of using `**` for units

https://github.com/key4hep/EDM4hep/blob/1cda5c55016eb760057607f7dba7dd591b859f59/edm4hep.yaml#L404

I don't have strong opinion whether to use `**` or `^` for units. But I think for `Chi^2` I'd prefer to rename it to "chi-square" so it's a plain text comment?

jmcarcell: I didn't want to be annoying but I agree with Andre on this one, `^2` I feel looks more physics-like and `**2` more computer-like, so I prefer ^2. At the same time, I think there was also a previous instance of **2 somewhere in the yaml file.

m-fila: No problem, you are never annoying :smile: `chi^2` is now `chi-squared`, units use `^` instead of `**`

andresailer: Thanks for indulging me!

andresailer: Yaml parser doesn't like `^`?

tmadlener: podio validation is at fault here:
https://github.com/AIDASoft/podio/blob/7a3598b187a014b764c3467c251b045057cf91f4/python/podio_gen/podio_config_reader.py#L39-L41

m-fila: Then we either revert to `**` or this waits for https://github.com/AIDASoft/podio/pull/755 (and becomes another reason to bump in #415)

andresailer: Go for `**`, I don't think my whims are worth bumping the required podio version for.

jmcarcell:
It's only because the description changed. I edited to include the whole message.

PR #425, opened 2025-03-31
Run pre-commit on Key4hep to have a recent enough clang-format

tmadlener:
BEGINRELEASENOTES
- Switch pre-commit workflow to run on Key4hep nightlies to have a recent enough version of `clang-format`.

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -22,6 +22,10 @@ jobs:
           STARTDIR=$(pwd)
           echo "::group::Build podio"
           cd ${STARTDIR}/podio
+          # Newer versions of git are more cautious around the github runner
+          # environment and without this git rev-parse --show-cdup in pre-commit
+          # fails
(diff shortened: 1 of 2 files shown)


m-fila:
With nightlies there is no need to `pip install`, right? https://github.com/key4hep/EDM4hep/blob/65d81e0c151438d1047f3cb29915d790acf5e9b1/.github/workflows/pre-commit.yml#L35

jmcarcell:
For pre-commit I opened this for a possible workflow https://github.com/key4hep/key4hep-dev-utils/pull/12. Originally started because EDM4hep needed to be fixed anyway, so let's fix all at once...

tmadlener:
I am not entirely sure if EDM4hep will be able to use the central workflow, as it builds podio@master on the fly. In any case this was more of a quick fix, so if it's possible with the central workflow, we can switch to that as well.

PR #426, opened 2025-04-23
Remove intermediate variables and make loading consistent in `__init__.py`

jmcarcell:
BEGINRELEASENOTES
- Remove intermediate variables and make loading libraries and files consistent in `__init__.py`

ENDRELEASENOTES

< 0 is bad (see https://root.cern.ch/doc/master/classTSystem.html#a6ff4c4c6cd8a75a199a30dd94359f205)  but 1 is returned when the library has already been loaded, which is not that bad but maybe we should know if that ever happens.

### DIFF
--- a/python/edm4hep/__init__.py
+++ b/python/edm4hep/__init__.py
@@ -5,28 +5,22 @@ import sys
 from .__version__ import __version__
 import ROOT
 
-res = ROOT.gSystem.Load("libedm4hepDict")
-if res < 0:
+if ROOT.gSystem.Load("libedm4hepDict") != 0:
     raise RuntimeError("Failed to load libedm4hepDict")
 
-res = ROOT.gSystem.Load("libedm4hepRDF")
-if res < 0:
+if ROOT.gSystem.Load("libedm4hepRDF") != 0:
     raise RuntimeError("Failed to load libedm4hepRDF")
 
-res = ROOT.gInterpreter.LoadFile("edm4hep/utils/kinematics.h")
-if res != 0:
+if ROOT.gInterpreter.LoadFile("edm4hep/utils/kinematics.h") != 0:
     raise RuntimeError("Failed to load kinematics.h")
 
-res = ROOT.gInterpreter.LoadFile("edm4hep/utils/dataframe.h")
-if res != 0:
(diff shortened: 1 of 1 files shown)


tmadlener (commented):
I am not sure, I would make it a hard-error if the library is already loaded. We will than have to make sure that we load them always in the right order, as otherwise a library might pull in another one as a dependency and then we will get a `1` and fail, when in reality it's not really a problem as ROOT keeps track of the libraries it has loaded.

m-fila:
I don't understand, in what situation requesting to load a library that is already loaded should be an error? Is it to internally sanitize how the libraries are loaded?

jmcarcell:
> I am not sure, I would make it a hard-error if the library is already loaded. We will than have to make sure that we load them always in the right order, as otherwise a library might pull in another one as a dependency and then we will get a `1` and fail, when in reality it's not really a problem as ROOT keeps track of the libraries it has loaded.

It is to be an error if a library is reloaded since only `!= 0` is checked.

> I don't understand, in what situation requesting to load a library that is already loaded should be an error? Is it to internally sanitize how the libraries are loaded?

I think in this case it's not too bad since there is probably something that is not working correctly. Like loading the edm4hep library and then loading edm4hep, I think that would trigger an error. In any case I think they should all be consistent, either `!= 0` always or `< 0` always.

jmcarcell:
So are we happy with this as it is? Note it only changes from `< 0` (error when loading) to `!= 0` for `libedm4hepDict` and `libedm4hepRDF`.

m-fila (approved):
I think I'd prefer to unify in the opposite direction so that `Load` is checked with `< 0`, but I guess it doesn't really matter - we had a mix and there were no issues, there shouldn't be any with consistent checks

tmadlener:
I think since we don't have duplicate loading at the moment we could switch to `!= 0` now. That makes it very easy to spot when we suddenly start to have duplicate loading and we can then decide whether it's a problem or whether it's our check that is to tight.

tmadlener (approved):

jmcarcell:
Yes, we can go back if needed. Maybe we find double loading of the dictionaries.

PR #430, opened 2025-05-15
Bump the required version of podio to 1.3.

### LINKED ISSUE #415: Bump minimum podio version to 1.3

jmcarcell:
BEGINRELEASENOTES
- Bump the required version of podio to 1.3. This is needed since ::typeName in link collections are only present in 1.3 and onwards. See https://github.com/key4hep/EDM4hep/issues/415

ENDRELEASENOTES

Closes #415

For versioning, EDM4hep 0.99.1 doesn't build with podio 1.3.


tmadlener (approved):
Waiting for tomorrows nightlies to pick up the podio tag.

PR #431, opened 2025-05-16
Change the EDM4hep version to 0.99.2 before creating a new tag

jmcarcell:
BEGINRELEASENOTES
- Change the EDM4hep version to 0.99.2 before creating a new tag

ENDRELEASENOTES

We are going to build a new release so we need a new tag for EDM4hep. I propose to use 0.99.2 instead of something bigger than the current one (like 0.999.0) that would be more confusing. There have been no breaking changes (https://github.com/key4hep/EDM4hep/compare/v00-99-01...main), or almost:
- https://github.com/key4hep/EDM4hep/pull/373
- https://github.com/key4hep/EDM4hep/pull/389
- https://github.com/key4hep/EDM4hep/pull/398

and I thin there should be nothing requiring 0.99.99. Then, after the tag, we would go back to 0.99.99.


tmadlener (commented):
This will be done automatically be the tagging script. Is there anything still missing for the tag from other PRs.

PR #433, opened 2025-06-10
Add a `stepLength` member to the `CaloHitContribution`

### LINKED ISSUE #432: Step length missing from CaloHitContribution
### Check duplicate issues.

- [ ] Checked for duplicates

### Goal

The original `lcio::SimCalorimeterHit` did have a `lengthCont` member for every contribution that has the step length. This seems to have been lost in translation and is needed for e.g. in the SDHCal digitization.

### Operating System and Version

Any

### compiler

Any

### The version of the key4hep stack

any

### Package Version

master

### Reproducer

none

### Additional context

_No response_

tmadlener:
BEGINRELEASENOTES
- Add a `stepLength` member to the `CaloHitContribution`
- Clarify the docstring for the `particle` relation of the `CaloHitContribution`

ENDRELEASENOTES

Fixes #432 

The unit is taken to be mm as that is how it is [filled for LCIO in DD4hep](https://github.com/AIDASoft/DD4hep/blob/e927f4947b6760c33df016a83a428054acd53d26/DDG4/lcio/LCIOConversions.cpp#L226).

Currently the RNTuple based backwards compatibility checks are failing, because RNTuple cannot yet do enough schema evolution to just add an empty member. Hence, it looks like the reading fails entirely.


tmadlener:
> Currently the RNTuple based backwards compatibility checks are failing,

At least for me locally. Maybe that is fixed with a newer version of ROOT as I don't see the same issue in CI.

tmadlener:
Indeed it looks like RNTuple that ships with ROOT 6.36 is able to read in the buffers that have been written with an older schema version. I have made the minimum ROOT version for the RNTuple backwards compatibility checks.

I don't think this is a controversial change, so I would merge this rather soon.

edm4hep.yaml
```diff
@@ -338,9 +338,10 @@ datatypes:
...
       - int32_t   PDG                        // PDG code of the shower particle that caused this contribution
-      - float energy [GeV]                    // energy of this contribution
-      - float time [ns]                     // time of this contribution
-      - edm4hep::Vector3f stepPosition [mm] // position of this energy deposition (step)
+      - float energy [GeV]                   // energy of this contribution
+      - float time [ns]                      // time of this contribution
+      - edm4hep::Vector3f stepPosition [mm]  // position of this energy deposition (step)
+      - float stepLength [mm]                // Geant4 step length for this contribution
     OneToOneRelations:
       - edm4hep::MCParticle particle     // primary MCParticle that caused the shower responsible for this contribution to the hit
```
BrieucF, line 346: While this is not directly part of this PR, shall we also change this comment "primary MCParticle that caused the shower responsible for this contribution to the hit" to "MCParticle that caused this hit contribution" because it looks like DDSIM does not ensure this (https://github.com/AIDASoft/DD4hep/blob/master/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L626-L630)? Or we change this behavior in DD4hep?

andresailer: This translates to the "primary" MCParticle, i.e. the particle that is still stored in the MCParticle record.

https://github.com/AIDASoft/DD4hep/blob/c7fc3e9ae29c1641c02c2aaf6b93b4fb321f0504/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L626-L627

BrieucF: "MCParticle that caused the hit. NB: if the MCParticle causing the hit is not kept in the MCParticle record, this will point to the first mother actually present in the output"?

tmadlener: Done with slight changes to your proposal. Let me know if you think this works.

tmadlener:
Given that this is rather uncontroversial, I will merge this later today.

PR #434, opened 2025-07-08
fix path to docs main page

m-fila:
BEGINRELEASENOTES
- Fixed blank documentation page with new doxygen version

ENDRELEASENOTES

It seems that with the current doxygen version the front page is blank. Setting full path to the file  fixes this.
The other pages weren't affected


jmcarcell:
It's now fixed :ok_hand:

PR #443, opened 2025-08-25
Use edm4hep.h in edm4hep2json to avoid hardcoding the types

jmcarcell:
Noticed after 64155dd, it would be better not to have to worry about types as
they change in this tool.

BEGINRELEASENOTES
- Use edm4hep.edm4hep in edm4hep2json to avoid hardcoding the types
- Remove the test `test_all_collections` that reads the .cxx file to check if
  all the collections are there (but a comment could trigger a false positive,
  for example). There is still a test that runs `edm4hep2json`.

ENDRELEASENOTES
  
I diffed a json file before and after these changes and I get the same result
(also for a file with a `UserDataCollection`).


m-fila (commented):
Very nice. I wondering if there was a reason not to do this or we just missed it earlier :slightly_smiling_face:

tmadlener (commented):
Very nice. I second the request from @m-fila to also have the user data types handled automatically (should be pretty straight forward with a `std::apply` over the `podio::SupportedUserDataTypes` tuple, I think)

m-fila:
How about defining `TypeList` in podio with all possible `UserDataCollection` types? I think it should have quite potential for reuse and be a slightly more handy than `SupportedUserDataTypes`

I mean something like this
```cpp
using UserDataCollectionList = decltype(std::apply(
    []<typename... Ts>(Ts...) {
        return utils::TypeList<UserDataCollection<Ts>...>{};
    },
    SupportedUserDataTypes{}
));
```

PR #445, opened 2025-09-04
Remove support for CPack and the unused `EDM4HEP.pc.in`

jmcarcell:
BEGINRELEASENOTES
- Remove support for CPack and the unused `EDM4HEP.pc.in`

ENDRELEASENOTES

I'm not aware of anyone using it and it hasn't receive any changes in 5 years so we can be decently sure no one is using it. It's also not an option in the Spack recipe.


tmadlener:
Fine for me. Given that this existed for quite some time. Maybe someone from the very beginning wants to object removing this? In that case at least some of the outdated links would need to be touched up.

PR #448, opened 2025-09-15
remove mentions of removed un-namespaced labels

m-fila:
BEGINRELEASENOTES
- Update comments referring to removed un-namespaced labels

ENDRELEASENOTES

Follow up to #440. The comments in yaml file were still referring to old labels (or even wrong names). Also reworded a bit to, hopefully, make it more clear what is a names and what is a variable.

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -401,7 +401,7 @@ datatypes:
       - float                phi [rad]       // Azimuthal angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
       - edm4hep::Vector3f    directionError [mm^2]  // covariance matrix of the direction
     VectorMembers:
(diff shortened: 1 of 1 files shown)


edm4hep.yaml
```diff
@@ -401,7 +401,7 @@ datatypes:
       - float                phi [rad]       // Azimuthal angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
       - edm4hep::Vector3f    directionError [mm^2]  // covariance matrix of the direction
     VectorMembers:
-      - float   shapeParameters      // shape parameters. This should be accompanied by a descriptive list of names in the shapeParameterNames collection level metadata, as a vector of strings with the same ordering
+      - float   shapeParameters      // shape parameters. The corresponding names of the shape parameters should be stored in the collection-level metadata collection identified by edm4hep::labels::ShapeParameterNames, as a vector of strings in the same order as the parameters.
```
tmadlener, line 404: Technically we don't really have collection-level metadata any longer. We have a dedicated naming convention for file-level metadata for "attaching" some parameters to a given collection. However, I don't know if that can be conveyed easily in some way.

PR #455, opened 2025-10-17
Clarify TrackerHitPlane member description

BrieucF:
BEGINRELEASENOTES
- Explicitly mention units and intended usage of TrackerHitPlane u, v, du and dv
ENDRELEASENOTES


tmadlener (approved):
To my understanding this says the same as the LCIO documentation.

PR #459, opened 2025-10-30
Cleanup setting of setting test environments and try to make it more robust

tmadlener:
BEGINRELEASENOTES
- Make the test environments for python based tests more focused on the current build
- Use fixtures to properly describe test dependencies
- Remove some of the duplication

ENDRELEASENOTES

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,26 +2,44 @@ IF(NOT BUILD_TESTING)
   RETURN()
 ENDIF()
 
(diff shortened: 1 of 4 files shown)


jmcarcell:
I still prefer having the loop, it's a one-time thing and then all tests can be added without any functions. The current one means every test has to use `add_edm4hep_test` to work. I see CI failed for [`778e76c` (#459)](https://github.com/key4hep/EDM4hep/pull/459/commits/778e76cda581757e1cb8dbf5e37bea859938de97) but I think this is now fixed (some exports that are needed for Key4hep@LCG were running always). At least I can run locally after running `k4_local_repo` as the CI will do now. I'll leave up to you to decide, since I don't care that much anyway.

PR #464, opened 2025-11-24
Store old schema versions explicitly

tmadlener:
BEGINRELEASENOTES
- Store (selected) old YAML definitions of the schema inside the `old_schemas` folder to allow for future code generation with schema evolution

ENDRELEASENOTES


tmadlener:
As discussed today during the meeting, I will merge this later today, unless there are any comments.

PR #467, opened 2025-11-28
Add more files for checking backwards compatibility

tmadlener:
BEGINRELEASENOTES
- Add more input files for backwards compatibility checks
- Fix a few version checks in the backwards compatibility checks to ensure they match the actual tags of EDM4hep

ENDRELEASENOTES

Some files have been created with an actual Key4hep release, others I have done with a local build.


tmadlener:
> Were half of the versions off by one in `test_EDM4hepFile.py`? It seems so 😕

Yes, I think it might have also been somewhat confusing during development because the tag versions weren't settled yet.

PR #468, opened 2025-11-28
Change type to use 32 bit integer in Quantity instead of 16 bit

tmadlener:
Otherwise we will just waste the padding between type and value, since value will be aligned on a 32 bit boundary again.

BEGINRELEASENOTES
- Make `Quantity::type` a 32 bit integer to not waste padding

ENDRELEASENOTES

As discussed / discovered at the [EDM4hep meeting on Nov 25](https://indico.cern.ch/event/1615736/)

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -206,7 +206,7 @@ components:
 
     edm4hep::Quantity:
       Members:
-        - int16_t type // flag identifying how to interpret the quantity
+        - int32_t type // flag identifying how to interpret the quantity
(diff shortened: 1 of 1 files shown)


tmadlener:
No, not really. @peremato simply spotted that we had an `int16_t` in there and was wondering whether that messes up the alignment. We then realized that the `value` is aligned on a 4 byte boundary again, and that we are simply wasting the padding here.

Re-ordering or trying to detect alignment issues would be a feature request for podio.

Issue #473, opened 2026-01-16
Downstream changes that need to land before we can tag 1.0


tmadlener:
- [x] https://github.com/AIDASoft/DD4hep/pull/1552
- [x] https://github.com/acts-project/acts/pull/4976
- [x] https://github.com/key4hep/k4FWCore/pull/371
- [x] https://github.com/key4hep/k4MarlinWrapper/pull/472
- [x] https://github.com/key4hep/k4EDM4hep2LcioConv/pull/130

PR #475, opened 2026-01-19
[wip] Bump project version to 1.0.0 (DO NOT MERGE)

tmadlener:
BEGINRELEASENOTES
- Thank you for writing the text to appear in the release notes. It will show up
  exactly as it appears between the two bold lines
- ...

ENDRELEASENOTES

Only here for checking whether tagging 1.0 breaks anything still

See also #473


tmadlener:
Looks like all downstream builds that are expected to build do so. I will make a tag later today.

PR #477, opened 2026-01-25
Fix a few issues in the vector utilities

jmcarcell:
BEGINRELEASENOTES
- `magnitude` isn't correct for 4D vectors. I've chosen to return a non-negative value which may be unexpected.
- `angleBetween` wasn't correctly implemented and was returning a value different than PI / 2 for orthogonal vectors. Avoid also dividing by zero.
- Add tests for the failures above

ENDRELEASENOTES

ROOT does throw an exception if you want to find the magnitude of a spacelike vector:
https://root.cern.ch/doc/master/classROOT_1_1Math_1_1LorentzVector.html#a26ec12d649ae172a505d2ef9a8a7f9b9

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,43 +13,43 @@ A generic event data model for future HEP collider experiments.
 
 | | | |
 |-|-|-|
-| [Vector4f](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L9)      | [Vector3f](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L39)      | [Vector3d](https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L67)     |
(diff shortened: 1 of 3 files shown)


jmcarcell:
I don't know, I think most of these utilities are not used. Well `magnitude` was returning a completely wrong number since it was implemented. I would like not to have to re implement them since they already exist in ROOT and, at least for <4D vectors, also in CLHEP.

tmadlener:
Maybe it would be more beneficial to have a (templated) `to` method that simply returns one of the wider spread vector types (in addition?). Something like the following for e.g. `Vector3D`

```cpp
template<typename T>
T to() const {
  return T{x, y, z};
}
```

Would still allow to specialize for types that don't simply construct from the coordinates.

jmcarcell:
> For unit vector `v` with `theta`, `phi` angles stored in `a` and `b` members the magnitude is 1, not `hypot(v.a, v.b, 0)` (we don't have to fix it here)

But we don't specify a coordinate system. In EDM4hep it just happens to be used as a unitary 3D vector, it could also be polar coordinates with a magnitude different than one. The better name would be "euclidean norm".

m-fila:
I think this would be equally misleading, as it makes assumption that the vector is written in orthonormal basis

Sorry for hijacking this thread with my 2D vector rant

PR #484, opened 2026-02-19
Allow C++26.

scott-snyder:
BEGINRELEASENOTES
- Allow compilation with C++26.
ENDRELEASENOTES


m-fila (commented):
`Key4hepConfig.cmake` is a managed file not to be overwritten. The source is here:
https://github.com/key4hep/key4hep-dev-utils/blob/main/defaults/cmake/Key4hepConfig.cmake

PR #497, opened 2026-04-21
Add operator== to ParticleIDMeta

jmcarcell:
BEGINRELEASENOTES
- Add operator== to ParticleIDMeta

ENDRELEASENOTES

Needed for https://github.com/key4hep/k4FWCore/pull/399, to compare if the existing metadata parameter is the same as the one we're trying to set.


tmadlener (commented):
Will this already deal with the `paramNames` member (`vector<string>`) correctly?

jmcarcell:
Yes, I think this will default to comparing each member.

tmadlener:
[Indeed](https://en.cppreference.com/cpp/container/vector/operator_cmp). I was not aware that was added in c++20. Nice :rocket:

tmadlener (approved):

jmcarcell:
This has always been there no? It only says it's constexpr since C++20.

tmadlener:
Yes, you are correct. I just never used this apparently...

PR #502, opened 2026-06-29
Build and install ArrowMapper if podio supports it

arnavdham:
BEGINRELEASENOTES

- Adds support for building and installing the ArrowMapper library when the installed version of podio was built with Arrow support (ENABLE_ARROW=ON). This follows the same pattern already used for SioBlocks.

ENDRELEASENOTES


test/CMakeLists.txt
```diff
@@ -111,7 +111,8 @@ find_package(HepPDT)
 
 if(HepMC3_FOUND AND HepPDT_FOUND )
   add_executable(edm4hep_testhepmc hepmc/edm4hep_testhepmc.cc)
-  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
+  target_include_directories(edm4hep_testhepmc BEFORE PRIVATE ${podio_INCLUDE_DIR})
```
jmcarcell: This looks unrelated, why is this needed?

jmcarcell:
Is it useful to have this at this time? Maybe it should be finished in podio first?

test/CMakeLists.txt
arnavdham: This is needed to prevent compilation failures in environments with older system-installed versions of podio.
The include paths for HepMC3 and HepPDT point to system directories (which also contain an older system-installed version of `podio/detail/Pythonizations.h`). That older version unconditionally includes `<Python.h>`, triggering a `Python.h file not found` error since the test target does not build with Python.

arnavdham:
> Is it useful to have this at this time? Maybe it should be finished in podio first?

@veprbl wanted to convert more complex edm4hep collections to arrow.

edm4hep/CMakeLists.txt
```diff
@@ -64,6 +64,13 @@ IF(TARGET edm4hepSioBlocks)
   list(APPEND EDM4HEP_INSTALL_LIBS edm4hepSioBlocks)
 ENDIF()
 
+PODIO_ADD_ARROW(edm4hep "${headers}" "${sources}")
```
jmcarcell: I'm not convinced about this, and maybe you want to deal with this now in https://github.com/AIDASoft/podio/pull/980. Have you tried to use it? If I'm not wrong, this call the macro that will check the current value of `ENABLE_ARROW`, which will never be ON because it's not an option in EDM4hep. Another downside is that this will fail for any podio version that doesn't have the last arrow PRs since PODIO_ADD_ARROW doesn't exist. It would be better to export something from podio and check if it's set; in that case add edm4hepArrow but don't fail otherwise.

test/CMakeLists.txt
jmcarcell: So it's unrelated, then it should go in another PR. Also it would be nice to have a reproducer.

arnavdham: I have reverted the test include changes from this PR and will submit them in a separate PR.

Regarding the reproducer, this compilation failure occurs when compiling the tests in an environment where:

An older version of podio is installed in a system/external directory (e.g. /opt/local/include or /usr/include), which happens to be the same prefix where HepMC3 or HepPDT is installed.
The project is being built against a newer version of podio (installed locally in a different path).
When target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR}) is called, it adds /opt/local/include to the search path before the linked target podio::podioRootIO's include path gets evaluated. Consequently, the compiler ends up finding the older system-installed podio/detail/Pythonizations.h instead of the local one:
```

In file included from .../edm4hep/test/hepmc/edm4hep_testhepmc.cc:7:
In file included from .../edm4hep/edm4hep/edm4hep/MCParticleCollection.h:15:
/opt/local/include/podio/detail/Pythonizations.h:5:10: fatal error: 'Python.h' file not found
    5 | #include <Python.h>
      |          ^~~~~~~~~~
1 error generated.
```

edm4hep/CMakeLists.txt
arnavdham: I have addressed this by updating AIDASoft/podio#980 to export a namespaced podio_ENABLE_ARROW status variable.

In this PR, I wrapped the PODIO_ADD_ARROW macro invocation in if(podio_ENABLE_ARROW). This avoids namespace conflicts and ensures the build won't fail when configuring against older versions of podio that do not support Arrow.

edm4hep/CMakeLists.txt
```diff
@@ -64,6 +64,15 @@ IF(TARGET edm4hepSioBlocks)
   list(APPEND EDM4HEP_INSTALL_LIBS edm4hepSioBlocks)
 ENDIF()
 
+if(podio_ENABLE_ARROW)
+  PODIO_ADD_ARROW(edm4hep "${headers}" "${sources}")
+  if(TARGET edm4hepArrow)
+    message(STATUS "Building and installing the Arrow since podio supports it")
```
jmcarcell:
```suggestion
    message(STATUS "Building and installing the Arrow component since podio supports it")
```

edm4hep/CMakeLists.txt
```diff
@@ -64,6 +64,15 @@ IF(TARGET edm4hepSioBlocks)
   list(APPEND EDM4HEP_INSTALL_LIBS edm4hepSioBlocks)
 ENDIF()
 
+if(podio_ENABLE_ARROW)
+  PODIO_ADD_ARROW(edm4hep "${headers}" "${sources}")
+  if(TARGET edm4hepArrow)
```
jmcarcell: Why this if? If we reached here `podio_ENABLE_ARROW` has to be true

jmcarcell:
Something seems not to work, the builds on CI on Alma 9 and Ubuntu 26 have podio built with Arrow but it doesn't seem to be triggering the build here, at least I can't find the message saying so.

test/CMakeLists.txt
jmcarcell: I see, can you copy this text into the text of https://github.com/key4hep/EDM4hep/pull/505?, then we can discuss it there

edm4hep/CMakeLists.txt
arnavdham: I was trying to keep it consistent with SIO/ROOT. Will remove this now as it is not requried.

arnavdham:
> Something seems not to work, the builds on CI on Alma 9 and Ubuntu 26 have podio built with Arrow but it doesn't seem to be triggering the build here, at least I can't find the message saying so.

The reason the Arrow component wasn't building is a bug in podio:

In `podioMacros.cmake`, the `PODIO_ADD_ARROW `macro returns early `if(NOT ENABLE_ARROW)`
However, podioConfig.cmake only exported `podio_ENABLE_ARROW `downstream, leaving `ENABLE_ARROW `undefined for edm4hep.
I think this should fix it: https://github.com/AIDASoft/podio/commit/94c90b5eaed73000b578e58120667a0ff36436eb
I can create a PR after you approve.

veprbl:
Why not just update https://github.com/AIDASoft/podio/blob/1b74f6fe06d8396e08e1dd109d25bde4e786c5d4/cmake/podioMacros.cmake#L351 ?

arnavdham:
> Why not just update https://github.com/AIDASoft/podio/blob/1b74f6fe06d8396e08e1dd109d25bde4e786c5d4/cmake/podioMacros.cmake#L351 ?

Because podio_enable_arrow is not defined during podio's own build, the `if(NOT podio_ENABLE_ARROW)` (I am assuming this is the change you want) will make it return early `TestDataModelArrow ` will never be created and later the test library link would fail.

This could work but isn't an elegant solution 
`if(NOT ENABLE_ARROW AND NOT podio_ENABLE_ARROW)`

@veprbl

veprbl:
I think this is quite conventional to do this double check.  I would also add `podio_ENABLE_SIO` for consistency.

arnavdham:
> I think this is quite conventional to do this double check. I would also add `podio_ENABLE_SIO` for consistency.

okay I can do that. But `SET(ENABLE_SIO @ENABLE_SIO@)` is already present so even if I remove my fix it will be inconsistent

arnavdham:
This should fix it: https://github.com/AIDASoft/podio/pull/990

@veprbl @jmcarcell

jmcarcell:
Looks good, this will be merged after https://github.com/AIDASoft/podio/pull/990 to make sure it works in the nightlies.

PR #505, opened 2026-07-15
Prioritize podio headers in testhepmc

arnavdham:
Re-applies the include path ordering fix to prioritize podio headers and avoid the system-shadowing compilation failure.
Discussed in https://github.com/key4hep/EDM4hep/pull/502

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -111,7 +111,8 @@ find_package(HepPDT)
 
 if(HepMC3_FOUND AND HepPDT_FOUND )
   add_executable(edm4hep_testhepmc hepmc/edm4hep_testhepmc.cc)
-  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR} )
+  target_include_directories(edm4hep_testhepmc BEFORE PRIVATE ${podio_INCLUDE_DIR})
+  target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR} ${HEPPDT_INCLUDE_DIR})
   target_link_libraries(edm4hep_testhepmc edm4hep podio::podioRootIO ${HEPPDT_LIBRARIES} ${HEPMC3_LIBRARIES})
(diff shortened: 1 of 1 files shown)


arnavdham:
@jmcarcell

jmcarcell:
I see the issue, but this is more generic than EDM4hep. If you are building something that depends on packages A, B, both installed globally, and you have a local A, then including B first always means that the headers from A are going to be picked up from the global and not the local installation. This will happen with any stack with views (for example, LCG stacks). I think that this is not a problem in EDM4hep, but rather the environment. The change you propose doesn't work if the package that is local would be HepMC3 (unlikely example, but it shows this is not a general fix, also the same issue would happen with HepPDT). Does setting `CPLUS_INCLUDE_PATH` to the include folder of your local podio help with this? Documentation says it goes after `-I`: https://gcc.gnu.org/onlinedocs/cpp/Environment-Variables.html.

arnavdham:
Regarding the reproducer, this compilation failure occurs when compiling the tests in an environment where:

An older version of podio is installed in a system/external directory (e.g. /opt/local/include or /usr/include), which happens to be the same prefix where HepMC3 or HepPDT is installed.
The project is being built against a newer version of podio (installed locally in a different path).
When target_include_directories(edm4hep_testhepmc PUBLIC ${HEPMC3_INCLUDE_DIR}) is called, it adds /opt/local/include to the search path before the linked target podio::podioRootIO's include path gets evaluated. Consequently, the compiler ends up finding the older system-installed podio/detail/Pythonizations.h instead of the local one:


```
In file included from .../edm4hep/test/hepmc/edm4hep_testhepmc.cc:7:
In file included from .../edm4hep/edm4hep/edm4hep/MCParticleCollection.h:15:
/opt/local/include/podio/detail/Pythonizations.h:5:10: fatal error: 'Python.h' file not found
    5 | #include <Python.h>
      |          ^~~~~~~~~~
1 error generated.
```

arnavdham:
Okay so i close this PR then? @jmcarcell

jmcarcell:
I'll close, but maybe try what I said in https://github.com/key4hep/EDM4hep/pull/505#issuecomment-5002979642 with `CPLUS_INCLUDE_PATH` to see if it works.

arnavdham:
Yes, it works.

PR #509, opened 2026-08-11
Update CMake configuration to use new PodioArrow library suffix

arnavdham:
Updates EDM4hep aliases and install rules to expect the edm4hepPodioArrow target. This reflects the macro changes in podio which renamed the generated suffix from Arrow to PodioArrow to avoid dynamic loading collisions.

Depends upon https://github.com/AIDASoft/podio/pull/999

BEGINRELEASENOTES
- **Fix:** Updated CMake targets to match the new `PodioArrow` suffix generated by podio's `PODIO_ADD_ARROW` macro.

ENDRELEASENOTES


tmadlener (approved):

tmadlener:
#510 should fix the remaining LCG based failures. Once those go through I will merge to make sure the Key4hep nightlies pass.

PR #510, opened 2026-08-13
Switch to nightlies stacks that exist for CI and update actions

tmadlener:
clang19 based stacks have been removed

BEGINRELEASENOTES
- Switch to `clang22` base stacks for LCG nightlies based CI workflows
- Switch to latest action versions and pin them via commit sha instead of just tags

ENDRELEASENOTES


andresailer:
clang22 uses c++23.

tmadlener:
That is an excellent catch. Thanks a lot.

PR #513, opened 2026-08-26
Fix units from directionError in edm4hep.yaml

selvaggi:
Direction error cannot be in mm^2.

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -402,7 +402,7 @@ datatypes:
       - edm4hep::CovMatrix3f positionError [mm^2]  // covariance matrix of the position
       - float                iTheta [rad]    // Polar angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
       - float                iPhi [rad]       // Azimuthal angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
-      - edm4hep::Vector3f    directionError   // covariance matrix of the direction
+      - edm4hep::Vector3f    directionError  [rad^2] // covariance matrix of the direction
     VectorMembers:
       - float   shapeParameters      // shape parameters. The corresponding names of the shape parameters should be stored in the collection named by edm4hep::labels::ShapeParameterNames in the file-level metadata, as a vector of strings in the same order as the parameters.
       - float   subdetectorEnergies  // energy observed in a particular subdetector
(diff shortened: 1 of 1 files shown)


tmadlener:
This should potentially also be an actual covariance matrix rather than just the diagonal(?) of it? Or for a quicker fix update the documentation to at least say that.

edm4hep.yaml
```diff
@@ -402,7 +402,7 @@ datatypes:
       - edm4hep::CovMatrix3f positionError [mm^2]  // covariance matrix of the position
       - float                iTheta [rad]    // Polar angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
       - float                iPhi [rad]       // Azimuthal angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
-      - edm4hep::Vector3f    directionError [mm^2]  // covariance matrix of the direction
+      - edm4hep::Vector3f    directionError   // covariance matrix of the direction
```
jmcarcell:
```suggestion
      - edm4hep::Vector3f    directionError  [rad^2] // covariance matrix of the direction
```
Since it is the covariance if `iPhi` and `iTheta`? Or at least it was originally (https://github.com/iLCSoft/LCIO/blob/master/src/aid/EVENT/Cluster.aid#L65 here it says so)

selvaggi: Indeed, this is probably intended to be a dimensionless 2x2 (theta, phi) covariance matrix

edm4hep.yaml
```diff
@@ -402,7 +402,7 @@ datatypes:
       - edm4hep::CovMatrix3f positionError [mm^2]  // covariance matrix of the position
       - float                iTheta [rad]    // Polar angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
       - float                iPhi [rad]       // Azimuthal angle of the cluster's intrinsic direction (used e.g. for vertexing). Not to be confused with the cluster position seen from IP
-      - edm4hep::Vector3f    directionError [mm^2]  // covariance matrix of the direction
+      - edm4hep::Vector3f    directionError  [rad^2] // covariance matrix of the direction
```
jmcarcell:
```suggestion
      - edm4hep::Vector3f    directionError [rad^2] // covariance matrix of the direction
```
Removing a space to be consistent with the other units

jmcarcell (approved):

edm4hep.yaml
tmadlener: I see, then `Vector3f` seems really off here. We would technially have a `CovMatrix2f` which would fit a bit better in principle.

How are people using this so far? Accessing the elements via `.x`, `.y`, and `.z` feels really wrong, so I assume they use `operator[]`?

selvaggi: agreed, would be better to have this as `CovMatrix2f`. Am about to use it for photon direction error as 
x = errorTheta^2
y = errorTheta*errorPhi 
z = errorPhi^2

tmadlener: Looks like Juan was quicker. But changing to `CovMatrix2f` might need some schema work in any case. That would then also allow us to explicitly assign the indices. Keeping track of that in https://github.com/key4hep/EDM4hep/issues/514

PR #515, opened 2026-08-28
Fix error prone retrieval of track states via location from Track

### LINKED ISSUE #511: Track::getTrackState is very easy to confuse with edm4hep::getTrackStates
`Track::getTrackState` is defined as

https://github.com/key4hep/EDM4hep/blob/f87de123bb1a56dfab7cdcb7f3cfb9dd51fc9313/edm4hep.yaml#L541-L549

It has essentially the same interface as `Track::getTrackStates` (which is generated from the `trackStates` vector member):

https://github.com/key4hep/EDM4hep/blob/f87de123bb1a56dfab7cdcb7f3cfb9dd51fc9313/edm4hep.yaml#L532

A simple typo can lead to a completely different (and most likely) wrong track state.

See also https://github.com/key4hep/k4GaudiPandora/pull/37

tmadlener:
BEGINRELEASENOTES
- Make it harder to use `Track::getTrackStates` with a location defined in `TrackState` by shifting the locations into an unscoped enum.
- The `Track::getTrackState(int)` overload has been deprecated and will be removed once all occurrences have been fixed

ENDRELEASENOTES

Fixes #511 

Turning the location into an unscoped enum keeps pretty much all of the existing behavior, but allows us to use a dedicated type for the `getTrackState` overload, while removing the overload for `getTrackStates(TrackState::Location)` entirely.

Having an unscoped enum has the advantages that it implicitly converts to int so existing code should keep working and also cppyy deals much better with this than a scoped enum as that doesn't implicitly convert and would require 

- [x] Check if / where this breaks downstream

### DIFF
--- a/edm4hep.yaml
+++ b/edm4hep.yaml
@@ -178,7 +178,7 @@ components:
     edm4hep::TrackState:
       Description: "Parametrized description of a particle track"
       Members:
-        - int32_t location // for use with At{Other|IP|FirstHit|LastHit|Calorimeter|Vertex}|LastLocation
+        - int32_t location // location of this track state, see edm4hep::TrackState::Location for the predefined values
         - float D0 // transverse impact parameter
         - float phi [rad] // azimuthal angle of the track at this location (i.e. not phi0)
         - float omega [1/mm] // is the signed curvature of the track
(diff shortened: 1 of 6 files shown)


edm4hep.yaml
```diff
@@ -537,17 +550,31 @@ datatypes:
...
+        /// Get the TrackState at a dedicated location
+        [[deprecated("Use getTrackState(edm4hep::TrackState::Location) instead")]]
+        std::optional<edm4hep::TrackState> getTrackState(std::int32_t location) const {
+          return getTrackState(static_cast<edm4hep::TrackState::Location>(location));
+        }
+
+        /// Deliberately deleted: getTrackStates(index) takes an *index* into the vector
+        /// of track states, not a location. Use getTrackState(location) to look up a
+        /// TrackState by its location.
+        /// See https://github.com/key4hep/EDM4hep/issues/511
```
tmadlener:
```suggestion
        /// TrackState by its location.
```

tmadlener:
I have checked this in a local stack and it seems to build without issues even if the currently deprecated `getTrackState(int)` is removed completely. We could consider dropping it entirely, but that might break code that is not part of the stack.

I have also verified that this would have lead to a build failure before the fixes https://github.com/key4hep/k4GaudiPandora/pull/37

test/test_ExtraCode.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
jmcarcell, line 1: Could this be made a separately registered test? Because its file name matches pytest's test_*.py pattern, it is collected by the bare pytest invocation in `check_complete_file`, which is very slow in CI.

tmadlener: You mean as a separately triggerable CTest registration?

tmadlener: I have split it into a separate `pytest` invocation and also actually split the whole suite into three actual test cases.

test/pytest.ini
```diff
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --ignore=test_rdf.py --ignore=tools --ignore=utils -p no:gaudi_fixtures
```
jmcarcell, line 2: Ideally this shouldn't need to be changed, it's easy to forget and tests can be run twice

jmcarcell:
https://github.com/key4hep/k4RecTracker/pull/98 seems to be needed

tmadlener:
Thanks. Curious that the k4RecTracker thing didn't show up in my local build. I thought I had checked this against a local stack. Sorry.

However, it also looks like it is already doing it's job.

test/pytest.ini
tmadlener: Yeah, maybe this should be done the other way around, i.e. exclude the long tests explicitly here and invoke them specifically and have all the rest be picked up by `pytest` automatically.

PR #516, opened 2026-08-31
Fix documentation of reco particle vertex relation & links

tmadlener:
Documentation still had a lot of outdated references of *associations* and the wrong name for the link type in general.

BEGINRELEASENOTES
- Fix the documentation for the `ReconstructedParticle` and `Vertex` relations and links

ENDRELEASENOTES


doc/VertexRecoParticleRecos.md
jmcarcell:
```suggestion
### Creating start vertex links
```
In at least another header that is changed but not here, maybe we should just use links everywhere and it would be simpler (there are still 3 or 4 places with associations in plain text).

tmadlener: Done. Looks like my search&replace was case sensitive

PR #518, opened 2026-09-02
Discover Python tests with pytest

jmcarcell:
Discover ordinary Python tests with one pytest command while reserving dedicated input-file tests.

BEGINRELEASENOTES
 - Discover ordinary Python tests with pytest.

ENDRELEASENOTES

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -101,6 +101,8 @@ IF(TARGET ROOT::ROOTDataFrame)
     FIXTURES_REQUIRED write_events_fixture
   )
 
+  add_edm4hep_test(py_test_rdf COMMAND python ${CMAKE_CURRENT_LIST_DIR}/test_rdf.py)
+  set_tests_properties(py_test_rdf PROPERTIES FIXTURES_REQUIRED write_events_fixture)
 endif()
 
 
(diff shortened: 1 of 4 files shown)


tmadlener (commented):
I am wondering whether we can safe some of the startup time related to `import ROOT` for the tests that do not read a file by having them run in one `pytest` invocation. I.e. doing sort of the opposite of what we did until now and explicitly exclude the `test_EDM4hepFile` tests via `pytest.ini` and have the rest auto-detect

jmcarcell:
Actually that stopped being a problem with newer versions of ROOT; latest stack on Alma 9 (in my local PC it's 31 ms):
```
% time python -c "import ROOT"
python -c "import ROOT"  0.05s user 0.02s system 67% cpu 0.103 total
```
EDM4hep takes more:
```
% time python -c "import edm4hep"
python -c "import edm4hep"  2.05s user 0.70s system 6% cpu 42.156 total
% time python -c "import edm4hep"
python -c "import edm4hep"  1.98s user 0.38s system 94% cpu 2.498 total
```
but by far it's the CVMFS time the one that is slow (like in the first try) in CI. Maybe it makes more sense to run them all together although I think the reporting is worse than if they are individual tests; I'll have a look.

tmadlener:
Ah good point about CVMFS. Since some of the tests run in parallel and they all wait for CVMFS to fill in the data, it probably doesn't gain us too much for CI. If checking can be done quickly, go for it. Otherwise it is what it is, and we should optimize for the local experience, IMO. (Which is what this PR already does).

jmcarcell:
Now it is done in a single test that shows up as `pytest` in CMake.

tmadlener (approved):

PR #519, opened 2026-09-05
Remove deprecated Cluster phi aliases

jmcarcell:
BEGINRELEASENOTES
- Remove the deprecated Cluster phi aliases. Under some circumstances ROOT reads the header files and emits deprecation warnings even when the aliases are not used.

ENDRELEASENOTES


jmcarcell:
Superseded by a PR from an upstream branch.

PR #521, opened 2026-09-13
build: fix hardcoded assumption that ROOT dictionaries are built

wdconinc:
When podio is built with ENABLE_ROOT=OFF, ROOT dictionaries are not built. This PR fixes the hardcoded assumptions in EDM4hep. See https://github.com/AIDASoft/podio/pull/1023.

A useful reviewer question is whether the better pattern is to gate on the `edm4hepDict` target (as in `edm4hep/CMakeLists.txt`) or on the `podio::podioRootIO` target (as in tests and tools). The reason for this bifurcation is that the `edm4hepDict` target is defined just before the gate in `edm4hep/CMakeLists.txt`, so it seemed more appropriate there.


jmcarcell:
Many of the tests depend on ROOT, they should be disabled.

edm4hep/CMakeLists.txt
```diff
@@ -54,9 +54,12 @@ if(NOT MERGE_RESULT EQUAL 0)
 endif()
 
 PODIO_ADD_ROOT_IO_DICT(edm4hepDict edm4hep "${headers}" ${MERGED_SELECTION_XML})
-add_library(EDM4HEP::edm4hepDict ALIAS edm4hepDict )
+list(APPEND EDM4HEP_INSTALL_LIBS edm4hep)
 
-list(APPEND EDM4HEP_INSTALL_LIBS edm4hep edm4hepDict)
+if(TARGET edm4hepDict)
```
jmcarcell, line 59: I'm wondering if it would be good to have a single option to enable or disable all the ROOT components; currently both `edm4hepDict` and `podioRootIO` are both checked. The two alternatives would be either looking at the beginning for one of the two and setting a variable and check that variable everywhere (with the benefit that it can be named as we want, such as `PODIO_WITH_ROOT`) or the same target everywhere.

wdconinc: Yeah, I added comments related to that in the PR description. To some extent they are serving different purposes. Maybe (ROOT v7? C++26 RTTI?) there will come a day when dictionaries are not needed but ROOT IO is still possible. The current approach tries to be a bit forward compatible with that future, but it does look like it is split.

jmcarcell:
I checked that none of the tests are removed in CI and that all the ROOT tests are removed with https://github.com/AIDASoft/podio/pull/1023, so this looks good to me.

edm4hep/CMakeLists.txt
jmcarcell: I don't have a strong opinion

jmcarcell (approved):
I checked that none of the tests are removed in CI and that all the ROOT tests are removed with https://github.com/AIDASoft/podio/pull/1023, so this looks good to me.

utils/CMakeLists.txt
```diff
@@ -23,22 +23,32 @@ target_include_directories(utils
...
   COMPONENT dev
   )
+if(TARGET ROOT::ROOTVecOps)
+  install(TARGETS edm4hepRDF
+    EXPORT EDM4HEPTargets
+    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT bin
+    RUNTIME DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
+    COMPONENT dev
+    )
+endif()
```
tmadlener, line 54: We could also use a list (like in `edm4hep/CMakeLists.txt`) so that there is only one install statement. It could also be a list that only holds the additional targets to install.

jmcarcell: Done

wdconinc:
I updated against main, and reverted the inadvertently added 0c383fa. That will be better as a separate PR since it changes behavior unrelated to this PR.

wdconinc:
@jmcarcell Are these test timeouts related? I don't think they are.

